Issue #7, opened 2016-01-02
'fisher install' fails due to mktemp on Mac OS X


tobywf:
There's two uses of `mktemp` in fisherman, which work fine on Linux but fail on Mac OS X because mktemp requires the `template` argument (see e.g. [stackoverflow](https://stackoverflow.com/questions/2792675/how-portable-is-mktemp1)).

I believe there's two easy solutions. One is to specify a prefix instead, the other is to see if `mktemp` failed:
```
mktemp -t fisher.XXXXXXXX
mktemp 2>/dev/null; or mktemp -t fisher
```
Either seems to work on Mac OS X (10.9 and 10.10), FreeBSD (9.3), Ubuntu (14.04), and Linux Mint (17.3).

tobywf:
Misread the man page, the FreeBSD I have (9.3) is actually fine with `mktemp`. So it's just OS X for now.

bucaran:
@tobywf Hi, what is the name of the plugin you could not install? I am on a OS X too.

tobywf:
This is on a totally new machine, with homebrew and fish installed:
```
fisher install oh-my-fish/theme-bobthefish
stderr: usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
Invalid redirection target: $tmp
~/.local/share/fisherman/functions/wait.fish (line 94):     fish -c "$commands" ^ $tmp &
                                                                                ^
in function 'wait'
        called on line 68 of file ~/.local/share/fisherman/functions/fisher_update.fish
        with parameter list '--spin=pipe --log=/Users/vagrant/.config/fisherman/cache/.debug_log curl -sS https://raw.githubusercontent.com/fisherman/fisher-index/master/INDEX > /Users/vagrant/.config/fisherman/ca
che/.26403'

in function 'fisher_update'
        called on line 103 of file ~/.local/share/fisherman/functions/fisher_search.fish
        with parameter list '--quiet --index'

in function 'fisher_search'
        called on line 3 of file ~/.local/share/fisherman/functions/fisher_install.fish
        with parameter list '--url=https://github.com/oh-my-fish/theme-bobthefish --field=name'

in command substitution
        called on line 0 of file ~/.local/share/fisherman/functions/fisher_install.fish

in function 'fisher_install'
        called on line 1 of file -
        with parameter list 'oh-my-fish/theme-bobthefish'

from sourcing file -
[...]

PR #8, opened 2016-01-02
Fix mktemp on Mac OS X

tobywf:
See #7, I've chosen the second approach to make `mktemp` work on Mac OS X.

### DIFF
--- a/functions/fisher_uninstall.fish
+++ b/functions/fisher_uninstall.fish
@@ -111,7 +111,7 @@ function fisher_uninstall -d "Disable / Uninstall Plugins"
             continue
         end
 
-        set -l tmp (mktemp)
+        set -l tmp (mktemp -t fisher.XXXXXXXX)
 
         if not sed -E '/^ *'(printf "%s|%s" $name $url | sed 's|/|\\\/|g'
         )'([ #].*)*$/d' < $file > $tmp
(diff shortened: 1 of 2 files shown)


functions/wait.fish
```diff
@@ -90,7 +90,7 @@ function wait -d "Run commands and wait with a spin"
             end
     end
 
-    set -l tmp (mktemp)
+    set -l tmp (mktemp -t fisher.XXXXXXXX)
```
bucaran: And here, `mktemp -t wait`.

functions/fisher_uninstall.fish
```diff
@@ -111,7 +111,7 @@ function fisher_uninstall -d "Disable / Uninstall Plugins"
             continue
         end
 
-        set -l tmp (mktemp)
+        set -l tmp (mktemp -t fisher.XXXXXXXX)
```
bucaran: @tobywf It seems that `mktemp -t fisher` is enough.

tobywf: On e.g. Ubuntu 14.04, the XXXXXX are required in a placeholder and get replaced with random characters (otherwise you get "mktemp: too few X's in template 'fisher'"). On FreeBSD and Mac OS X, they don't get replaced but they pose no problem.

bucaran: You are right. I have fixed it in 0.2.0 which I should be pushing shortly. :smile:

bucaran:
Closing in favor of #18 

Thanks for your contribution @tobywf! Keep them coming ^^/

PR #26, opened 2016-01-07
We have moved into slack room from gitter

simnalamburt:
We have moved into slack room from gitter


bucaran:
:tropical_fish: Thanks!

PR #80, opened 2016-02-07
Use command when calling non-builtins

### LINKED ISSUE #79: use fully qualified paths of `cp`, `mv`, etc. or `command` before invoking them
I have common commands like `cp`, `mv`, etc. aliased to `cp -i`, `mv -i`, etc.

When I use fisherman, I see `overwrite /Users/jab/.config/fisherman/fishfile? (y/n [n]) not overwritten`.

I assume this is caused by fisherman referring to a command like cp as just `cp`, rather than as `/bin/cp` or `command cp`. This can cause problems if the user has the command aliased to pass other arguments (such as interactive mode, as above), since fisherman is assuming `cp` is the same thing as `/bin/cp`, it isn't. If so, any interest in fixing this? I think well-behaved shell scripts are supposed to use fully-qualified paths or use `command` to avoid this problem.

daenney:
By using `command` we ensure that we always end up calling the external
binary and don't get caught by user aliasses, functions or other
magic.

Closes #79

This is a find-replace powered fix, I might have missed something. I purposefully didn't alter the tests so that we could catch any discrepancies.

### DIFF
--- a/functions/__fisher_index_update.fish
+++ b/functions/__fisher_index_update.fish
@@ -18,9 +18,9 @@ function __fisher_index_update -a timeout
     end
 
     if not curl --max-time $timeout -sS "$url" > $index
-        rm -f $index
+        command rm -f $index
         return 1
     end
 
-    mv -f $index $fisher_cache/.index
+    command mv -f $index $fisher_cache/.index
 end
--- a/functions/__fisher_plugin_enable.fish
+++ b/functions/__fisher_plugin_enable.fish
@@ -39,7 +39,7 @@ function __fisher_plugin_enable -a plugin path
             case --uninstall
             case \*
                 if test "$class" = --man
-                    mkdir -p (dirname $fisher_config/$target)
+                    command mkdir -p (dirname $fisher_config/$target)
                 end
 
                 __fisher_plugin_link $link $source $fisher_config/$target
--- a/functions/__fisher_file_remove.fish
+++ b/functions/__fisher_file_remove.fish
@@ -8,6 +8,6 @@ function __fisher_file_remove -a plugin file
 
         sed -E "$pattern" < $file > $file.tmp
 
-        mv $file.tmp $file
+        command mv $file.tmp $file
     end
 end
--- a/functions/__fisher_key_bindings_disable.fish
+++ b/functions/__fisher_key_bindings_disable.fish
@@ -7,10 +7,10 @@ function __fisher_key_bindings_disable -a plugin user_key_bindings
         > $fisher_key_bindings.tmp \
         < $fisher_key_bindings
 
-    mv -f $fisher_key_bindings.tmp $fisher_key_bindings
+    command mv -f $fisher_key_bindings.tmp $fisher_key_bindings
 
     if test ! -s $fisher_key_bindings
         sed -i.tmp '/__fisher_key_bindings/d' $user_key_bindings
-        rm -f $user_key_bindings.tmp
+        command rm -f $user_key_bindings.tmp
     end
 end
--- a/functions/__fisher_key_bindings_enable.fish
+++ b/functions/__fisher_key_bindings_enable.fish
@@ -2,7 +2,7 @@ function __fisher_key_bindings_enable -a plugin user_key_bindings
     __fisher_key_bindings_update $plugin >> $fisher_key_bindings
 
     if test ! -s $user_key_bindings
-        mkdir -p (dirname $user_key_bindings)
+        command mkdir -p (dirname $user_key_bindings)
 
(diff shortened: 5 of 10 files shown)


functions/__fisher_index_update.fish
```diff
@@ -10,17 +10,17 @@ function __fisher_index_update -a timeout
...
 
-        # Force the server to return the latest copy of the index using a fake query string.
+        # Force the server to return the lacommand test copy of the index using a fake query string.
 
         set url https://raw.githubusercontent.com/fisherman/fisher-index/master/index\?(date +%s)
     end
 
-    if not curl --max-time $timeout -sS "$url" > $index
-        rm -f $index
+    if not command curl --max-time $timeout -sS "$url" > $index
```
bucaran: Unfortunately we won't be able to `command curl` yet, maybe in the future, but currently, we need to use the default `curl` in order to run some tests that basically "mock" `curl`.

functions/__fisher_path_from_url.fish
```diff
@@ -8,8 +8,8 @@ function __fisher_path_from_url -a url
 
     for file in $fisher_cache/*
         switch "$url"
-            case (git -C $file ls-remote --get-url)
-                printf "%s\n" $file
+            case (command git -C $file ls-remote --get-url)
```
bucaran: The same that goes for `curl`, goes for `git`, unfortunately.

functions/__fisher_path_make.fish
```diff
@@ -1,13 +1,13 @@
 function __fisher_path_make -a path
-    if test -s $path/Makefile -o -s $path/makefile
-        pushd $path
+    if command test -s $path/Makefile -o -s $path/makefile
+        command pushd $path
```
bucaran: `pusdh` is a `fish` function, so we don't need this one.

functions/__fisher_plugin_disable.fish
```diff
@@ -1,5 +1,5 @@
 function __fisher_plugin_disable -a plugin path
-    __fisher_plugin_walk "$plugin" "$path" | while read -l class source target name
+    __fisher_plugin_walk "$plugin" "$path" | while command read -l class source target name
```
bucaran: `read` is a builtin, so it would be `builtin read`, however, let's stick to the most basic commands for now.

functions/__fisher_plugin_enable.fish
```diff
@@ -17,7 +17,7 @@ function __fisher_plugin_enable -a plugin path
 
     set -l link -f
 
-    if test -L $path
+    if command test -L $path
```
bucaran: `test` is also a builtin.

functions/__fisher_plugin_disable.fish
daenney: Ah, I thought it was calling out to `/usr/bin/read`, I'll fix that right up.

functions/__fisher_xdg.fish
```diff
@@ -5,21 +5,21 @@ function __fisher_xdg -a dir
 
     switch "$dir"
         case --config{,-home}
-            if set -q XDG_CONFIG_HOME
+            if command set -q XDG_CONFIG_HOME
```
bucaran: Oops :wink:

functions/fisher_help.fish
```diff
@@ -104,14 +104,14 @@ function fisher_help -d "Show Help"
                                 print ""
                                 exit
                             }'
-                    end | column -ts\t
+                    end | command column -ts\t
```
bucaran: Haha! Maybe this is too much?

daenney: Probably. But consistency! :smile:.

bucaran:
Thanks! @daenney I'll be monitoring the tests and see what is the outcome.

For the real deal, I'd like to begin simply by replacing only the following commonly aliased commands:

* cp
* ln
* rm

We may end up using `command` for more commands in the future, but right now, while I honestly can't make an accurate assesment regarding performance, my gut tells me that this will slow things down a bit.

daenney:
So far doing these calls a few hundred times over with `time` doesn't reveal any noticeable difference.

bucaran:
That's good. Let's make a list of what commands we are going to `command` patch and let's track it in #79.

bucaran:
@daenney Thanks! :white_check_mark:

daenney:
@bucaran I can't find any `cp` calls in `functions/`, only in `tests/`.

bucaran:
Gotcha. Alright, I'll test this on my machine and hopefuly get it merged tomorrow! Thanks @daenney.

Issue #129, opened 2016-02-26
Beautify man pages


pickfire:
- [x] Beautify those ascii tree diagram
- [x] [Words to avoid](https://github.com/nicholaides/words-to-avoid.vim) in those articles
- [x] Misc fixes / forgotten updates

bucaran:
+1 for words to avoid and I am sure I can think of more things.

One more thing I am trying actively to avoid is using adverbs.

As for the ascii tree diagram, I recall my `man` would not display unicode characters, so I desisted.

pickfire:
> As for the ascii tree diagram, I recall my man would not display unicode characters, so I desisted.

Of course, I always know that and I will make that my first policy, I just want to say that the ascii tree diagram still looks a bit ugly and some part needs some fix.

> One more thing I am trying actively to avoid is using adverbs.

I am not sure what you mean here, one more thing that I am doing is to make those sentences as short as possible as long as the original meanings remain.

bucaran:
Cool. Go for it.

bucaran:
@pickfire Now that Fisherman API is essentially locked, I will go over the documentation and rewrite and fix as much as possible too. In addition, I will simplify the Wiki, re-shoot gifs with the current Fisherman and consolidate as much as possible in preparation for 1.0.

pickfire:
I need to go through all the docs, give me some time. At the moment, you can work on the wiki.

bucaran:
@pickfire If you haven't started, then leave that up to me, and you can review it after I push the new version. I will be doing this from now, over the next few hours. I would love if you can fix anything after that.

pickfire:
I have started since last night, assign this to me and the other parallel thing (haven't started).

bucaran:
Well, the thing is, I will remove some of the docs and consolidate them into one or two documents. I wrote too much docs to begin with.

PR #131, opened 2016-02-27
[#129] Beautify man pages

pickfire:
Sorry for the late commit, I don't have the time to read through all the docs, I just skim though it and fix everything quickly.

### DIFF
--- a/man/man1/fisher-help.md
+++ b/man/man1/fisher-help.md
@@ -3,7 +3,7 @@ fisher-help(1) -- Show Help
 
 ## SYNOPSIS
 
-fisher `help` [*keyword*] [`--all`] [`--guides`] [`--usage`[=*command*]] [`--help`] <br>
+fisher `help` [*keyword*] [`--all`] [`--guides`] [`--usage`[=*command*]] [`--help`]<br>
 
 ## USAGE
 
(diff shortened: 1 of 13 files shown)


bucaran:
@pickfire Thanks so much! Mind you, I will end up merging some of these documents later and simplifying even more, but this should

bucaran:
@pickfire I merged locally and will push with your commits later today. Please don't add anything else here.

You are free to take this on after I merged the new docs. A lot of docs were removed and simplified into `fisher.1`.

pickfire:
Ah, I am currently not working on fisherman, I don't have anything interested to do in fisherman except for the support of parallelism.

bucaran:
@pickfire Sure, give that a shot. I would love to see what some benchmark results.

pickfire:
@bucaran, I didn't do any benchmarks, I can even feel that it is slower (my eyes and brain) so it must means that there is about one or two seconds slower for `parallel`.

bucaran:
So, why is the point of implementing a feature that is slower?

pickfire:
That's why I am asking for your opinion, whether `xargs` or `parallel`. I suggest `xargs` for this.

bucaran:
I say `xargs` too.

pickfire:
@bucaran Okay, now I am thinking about adding it to fisherman, give me some time.

PR #173, opened 2016-03-27
make Fisherman vars transparent

orangecms:
This allows a user to include Fisherman through just a single line in their config, they would just need to `source path/to/fisherman/config.fish`. Gaining this transparency, at the same time they can also override the vars if they wish.

As a plus, Fisherman should check for the necessary config directory to be present and create it otherwise. This way people do not need to do it themselves during installation. At the same time, that allows for easy installation to their desires, e.g. to a global path, w/o reverse-engineering the Makefile or installation script. That is also similar to Vundle, dein, and other plugin managers.

### DIFF
--- a/config.fish
+++ b/config.fish
@@ -1,3 +1,23 @@
+# If fisher_home is not provided, use the directory of this file:
+set -q fisher_home; or set -g fisher_home (dirname (status -f))
+
+# If fisher_config is not provided, set it up:
+set -q fisher_config
+or if test ! -z "$XDG_CONFIG_HOME"
+    set -g fisher_config $XDG_CONFIG_HOME/fisherman
+else
+    # if XDG home is unset or empty, use ~/.config as a fallback
+    set -g fisher_config ~/.config/fisherman
+end
+
+# To not break fish, we need to make sure the config dir exists:
+if test ! -d "$fisher_config"
+    echo "Setting up Fisherman config dir..."
+    mkdir -p "$fisher_config/cache"
+    touch "$fisher_config/fishfile"
+    echo "Done."
+end
+
 set -g fisher_file $fisher_config/fishfile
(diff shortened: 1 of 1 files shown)


bucaran:
I'd rather tell users to configure their config.fish to define fisher_* vars they need and keep our `config.fish` simple and fast.

orangecms:
The vars should still be checked imho. If the config dir is missing, it'll break the system, and for the unexperienced average user who set fish as their default shell, thus their whole system. That doesn't sound too nice to me.

bucaran:
> If the config dir is missing, it'll break the system, and for the unexperienced average user who set fish as their default shell, thus their whole system. That doesn't sound too nice to me.

The average user will use the magic install, so this will never happen.

The advanced user probably knows what they are doing and can figure it out. Plus, the steps will be thoroughly documented too.

orangecms:
That's about the discussion we had in Slack. There are average users who would like to install software through their system's package manager as well, which is still the de facto way of installation. And I think there is no excuse for not providing sanity checks. Meaning loss of a millisec is a bit weird, that doesn't really reduce speed. I mean, it would also be faster if the file system wrote without parity and stuff, but it leads to errors.

bucaran:
To be clear, I am *not* against the sanity checks, we can have them, we should have them, literally, anywhere, either inside `fisher.fish`, or via another script that would runs the first time fisherman is installed.

Now, the only place where I won't have the sanity checks is in our `config.fish`. I just don't want to run that code every time I create a new session.

orangecms:
Then how about offering a bootstrapping script that writes to the user's fish config, something like a `/usr/share/fisherman/bootstrap.fish`?

bucaran:
@orangecms I am going to close this in favor of the make PR, let's keep track of all of this over there.

PR #174, opened 2016-03-27
Usual Makefile

orangecms:
A `Makefile` usually offers two things, allowing for easy packaging and installation:
- a `make` step which creates a build (not yet necessary for Fisherman)
- a `make install` step which copies the resulting files to a given `$PREFIX` directory

So you could run e.g. `PREFIX=/usr/local make && make install`.

The current Makefile, however, manipulates a user's `config.fish` without their consent and doesn't create something like a build. This means that packagers cannot use it at all and have to find their own way of copying the files, figuring out which are needed and where to put them, etc..

I haven't yet really investigated the `make test` step, but it should finally be offered as well. It should be possible to run it easily, but currently requires `fishtape` to be present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+sharepath = $(DESTDIR)/usr/share
+fishpath = $(sharepath)/fish
+fisherpath = $(sharepath)/fisherman
+manpath = $(sharepath)/man
+
+.PHONY: all
+all:
+
+.PHONY: install
+install:
+	# prepare directories because install -Dt won't do it on the CI server
+	mkdir -p "$(fisherpath)/completions" "$(fisherpath)/functions"
+	# the main config, which can be sourced by the user
+	install -Dm 644 config-fisherman.fish "$(fishpath)/config-fisherman.fish"
+	install -Dm 644 config.fish "$(fisherpath)/config.fish"
+	install -Dm 644 config-fisherman.fish "$(fishpath)/config-fisherman.fish"
+	# completions
+	install -Dt "$(fisherpath)/completions" completions/*
+	# functions
+	install -Dt "$(fisherpath)/functions" functions/*
+	# man pages, README and LICENSE
+	install -Dt "$(manpath)/man1" man/man1/*
+
+.PHONY: clean
+clean:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,8 @@ before_install:
     - sudo apt-get update
     - sudo apt-get -y install fish
 script:
-    - bin/install
-    - fish -c "fisher i fishery/fishtape; fishtape test/*.fish"
+    - sudo make install
+    - mkdir -p ~/.config/fish
+    - echo "source /usr/share/fish/config-fisherman.fish" >> ~/.config/fish/config.fish
+    - fish -c "fisher i fishtape"
+    - fish -c "fishtape test/*.fish"
--- a/README.md
+++ b/README.md
@@ -27,6 +27,19 @@
 curl -sL get.fisherman.sh | fish
 ```
 
+## Packaging
+
+Fisherman supports the [DESTDIR convention](https://www.gnu.org/prep/standards/html_node/DESTDIR.html). Using the Makefile (`DESTDIR` should be adjusted with respect to your OS/distribution or simply not provided):
+```bash
+make install DESTDIR=/path/to/stage
+```
+
+### Usage prerequisites
+
+In order for users who install the package to actually use Fisherman, they need to source the provided `/usr/share/fish/config-fisherman.git`. In your package, you should thus provide a hint such as the following:
+
+Please run `echo "source /usr/share/fish/config-fisherman.git" >> ~/.config/fish/config.fish` to enable Fisherman for your current user.
+
 ## Usage
 
 Install plugins.
--- a/config-fisherman.fish
+++ b/config-fisherman.fish
@@ -1,0 +1,21 @@
+# If fisher_home is not provided, use the global directory:
+set -q fisher_home; or set -g fisher_home (dirname (status -f))/../fisherman
+
+# If fisher_config is not provided, set it up:
+set -q fisher_config
+or if test ! -z "$XDG_CONFIG_HOME"
+    set -g fisher_config $XDG_CONFIG_HOME/fisherman
+else
+    # if XDG home is unset or empty, use ~/.config as a fallback
+    set -g fisher_config ~/.config/fisherman
+end
+
+# To not break fish, we need to make sure the config dir exists:
+if test ! -d "$fisher_config"
+    echo "Setting up Fisherman config dir..."
+    mkdir -p "$fisher_config/cache"
+    touch "$fisher_config/fishfile"
+    echo "Done."
+end
+
+source "$fisher_home/config.fish"


bucaran:
So basically no makefile? I like this.

A lot of other places will need to be updated before this can be merged though.

orangecms:
What do you mean by "no makefile"?

orangecms:
http://www.network-theory.co.uk/docs/gccintro/gccintro_16.html

Here's what a Makefile should be used for and how. Tests should run in `make check`, apparently.

bucaran:
> What do you mean by "no makefile"?

I mean a very short makefile.

orangecms:
Yea it's basically short for now because not many files need installation and there is no build procedure involved yet.

bucaran:
@jethrokuan Comments :)

bucaran:
@orangecms Regarding the PR that modifies the `config.fish`, I think that one should be discussed last. First let's decide on the new install mechanism, basically let's work on this PR and then see what needs to be done in terms of the `config.fish`.

orangecms:
In summary, I suggest differiantiating between installing Fisherman in the system and adding it to the user's config. The latter can/should not be covered by a system's package manager, unless we write to fish's config in /etc or /usr/etc or the new conf.d directory, which can be done in a post-transaction hook. But since that is run by the system's package manager's user, it would mean mixing things again and would require a user to modify their config file in order not to break their system. That's why I would do the sanity checks for the existence of the essential config dir.

Makefile
```diff
@@ -1,62 +1,14 @@
...
-		ronn --manual=fisherman --roff $? 1>&2 2> /dev/null;\
-	fi;\
+sharepath = "$(PREFIX)/share"
+fisherpath = "$(sharepath)/fisherman"
+manpath = "$(sharepath)/man"
+
+.PHONY: install
+install:
+	# the main config, which can be sourced by the user
+	install -Dm 644 config.fish "$(fisherpath)/config.fish"
```
jethrokuan: What we don't want to do is to override `config.fish` the user might have some stuff preset.

jethrokuan:
:+1: on the `conf.d` method in the future.

While the one line make is great and all, I think it's best we stick to convention and name it `make install`. Experienced users will `make`, and the average user will use the curl installer.

I agree on not doing checks in `config.fish`. There's no need to compromise speed here.

We could:
Have `fisher doctor`, which checks for missing variables, outdated fisherman install and broken fishfile etc.

orangecms:
That would only work as long as fish works though, since fisher is not a standalone tool. We should discuss that in another issue.

bucaran:
Ping @orangecms 

Do you think we can have this PR redone and updated as previous discussed?

Makefile
```diff
@@ -1,62 +1,14 @@
...
-MAN := $(FISHER_HOME)/man
-MAN1 := $(wildcard $(MAN)/man1/*.md)
-DOCS := $(MAN1:%.md=%.1)
-
-INDEX := $(FISHER_CACHE)/.index
-VERSION = `cat $(FISHER_HOME)/VERSION`
-
-.PHONY: all test flush uninstall release
-
-all: $(FISH_CONFIG) $(FISHER_CACHE) $(FISHER_FILE) $(DOCS)
```
pickfire: Where did `make all` went? If it isn't here, isn't `make install` just the same as `make`?

pickfire:
My simple and featureful makefile: https://github.com/pickfire/www/blob/master/Makefile

orangecms:
No. `make` has to create a build, but we don't have build steps (for now). `make install` must copy the necessary files with certain access rights to a given prefix directory. These are *NIX / GNU conventions. @bucaran I have rebased w/o an issue (as I expected), but I haven't yet looked into autotools for the configuration. This is where the prefix should be determined iirc.

bucaran:
@orangecms Thanks. This PR should also contain changes in the configuration, and docs updates.

jethrokuan:
also, change the travis build for your PR, and see if you can make tests pass

orangecms:
Yep, got that. ;)

bucaran:
Ping @orangecms

Hey man, would love to see where this goes or have at least a plan.

orangecms:
The plan is to check out autotools und figure whether and how they can help to obtain the necessary structure, which might result in an actual build process where paths to the directories would need adjustment. The second hurdle for me is modifying the CI scripts for acceptance. I don't have much experience with OS-independent builds yet. If anyone knows about autoconf/`.ac`-files, that would really help.

orangecms:
big kudos to @pickfire for the `addprefix`, I'll add that as well!

orangecms:
I will read http://www.gnu.org/software/make/manual/make.html#Directory-Variables later and see how to do it properly. :)

orangecms:
@bucaran Now the only missing piece is adding a hint in the README to `source $prefix/share/fish/config-fisherman.fish` (or create the dirs manually, set the env vars and source fisherman's `$prefix/share/fisherman/config.fish`) :)

bucaran:
@orangecms Great, do you have a PR ready?

@denji We should coordinate the update with the release of a brew package as well.

@jethrokuan Comments?

jethrokuan:
@bucaran with fisherman merging into fin I can imagine the make installation might simplify as well.

But other than that I'm all for this change, and it's one that we really need.

bucaran:
Plus @dalln1 is done with fin-git too, so it will be basically just modifying the name and other details.

@orangecms Can you pair up with @dalln1?

dalln1:
As far as I understand it I can not change the name, I've to delete fin-git on the AUR and we have to modify the fisherman-git on the AUR for the upcoming release of fisherman.

bucaran:
@dalln1 Thanks. I should be done with this soon.

dalln1:
@bucaran Perfect. I've just seen that there is also the possibility to merge two packages on the AUR. But not sure which way will be easier. I think I will wait what @orangecms has to say about this, because he is more experienced than me with the arch packaging system.

bucaran:
Oh, I am pretty sure @orangecms will have a lot to say about all this 😟 hehe

orangecms:
@bucaran the PR is right here, this discussion belongs to it ;)
@dalln1 Let's have a chat about the packages in the Slack team if you're in there.

bucaran:
@orangecms Thanks! Now that 2.0 is out there, what should we do with this?

orangecms:
I dunno, I can rewrite it for the current setup.

bucaran:
Yes, please. 🙏

bucaran:
@orangecms With the new AUR package we don't need a makefile anymore, so can we close this?

orangecms:
I would suggest to maintain a legacy branch for people who want it (e.g. for the search feature). A Makefile can still be writte  for 2.x, its release is unrelated to offering a Makefile.

bucaran:
@orangecms 
>  would suggest to maintain a legacy branch for people who want it

I don't feel like maintaining another branch for something I will never use, plus, I hope to bring a little bit of the search functionality back into fisherman at some point without modifying the cli, for example `fisher ls --remote` or create a separate plugin for it.

bucaran:
Closing as not needed now that we have the AUR pkg and installing fisherman is super easy.

Thanks for all the work @orangecms.

PR #175, opened 2016-03-29
For some reason a \ line continuation followed by a | at the beginning

BretJohnson:
of the next line has problems, at least on my machine running cygwin.
It produced an error like this:
~/.local/share/fisherman/functions/__fisher_plugin_validate.fish (line
42):                 | sed -E "
^ from sourcing file
~/.local/share/fisherman/functions/__fisher_plugin_validate.fish
called on standard input

This change simply eliminates the \ line continuation and instead
concatenates with the next line.  It seems simple enough and avoids
this problem, making it work on cygwin.

### DIFF
--- a/functions/__fisher_plugin_validate.fish
+++ b/functions/__fisher_plugin_validate.fish
@@ -38,8 +38,7 @@ function __fisher_plugin_validate -a plugin
                 return 1
             end
 
-            printf "%s\n" "$plugin" \
-                | sed -E "
+            printf "%s\n" "$plugin" | sed -E "
                     s|^gh[:/]+|https://github.com/|
                     s|^gl[:/]+|https://gitlab.com/|
                     s|^bb[:/]+|https://bitbucket.org/|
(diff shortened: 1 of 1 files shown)


bucaran:
Thanks @BretJohnson.

It does seem simple enough so I'm just going to merge.

Ping @orangecms @jethrokuan

jethrokuan:
do you think it could be because of how windows handle newlines?

BretJohnson:
@jethrokuan Yes, I think that's it.  I reinstalled fisherman to reproduce the problem.  Then I tried changing the fish script to have Unix-style (LF) line endings, via this:
Bret@BABYBLUE ~/inst> dos2unix ~/.local/share/fisherman/functions/__fisher_plugin_validate.fish
dos2unix: converting file /c/Dropbox/home/.local/share/fisherman/functions/__fisher_plugin_validate.fish to Unix format...

And indeed that fixed it it.  For cygwin, normally shell scripts (at least bash scripts) have Unix-style line endings, even on Windows.  Bash scripts won't work at all if they don't.  Fish scripts seem more resistant to that, working even with CR+LF line endings, except for this scenario.  So this may ultimately be a fish bug, handling the \ at end of line, on Windows in files with CR+LF line endings.

bucaran:
@BretJohnson Great insight, and thanks for asking this question @jethrokuan.

Bug in fish or not, this little patch fixes Fisherman for @BretJohnson so let's merge.

jethrokuan:
:+1: Thanks for all the effort @BretJohnson , hopefully we can bring this up to the fish devs.

orangecms:
Oh nice find, thanks @BretJohnson :)

PR #183, opened 2016-04-01
Update README.md

xrain0610:
Add the chinese support notice

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
     <br>
 </h1>
 
+**[Fisherman]** 现在提供中文支持,如果你有任何问题,可以以中文提交ISSUE,我们会尽快解决,感谢您的支持!
 **[Fisherman]** is a blazing fast, modern plugin manager for [fish].
 
 ```fish


bucaran:
@pickfire Ping.

jethrokuan:
@bucaran this just means we'll provide support for github issues in chinese.

bucaran:
@xrain0610 @jethrokuan Great. Let's add a missing new line between the Chinese and English lines.

README.md
```diff
@@ -14,6 +14,7 @@
     <br>
 </h1>
 
+**[Fisherman]** 现在提供中文支持,如果你有任何问题,可以以中文提交ISSUE,我们会尽快解决,感谢您的支持!
```
pickfire: I suggest to use `若有疑问` instead of `如果你有任何问题`.

pickfire:
@bucaran Nice. I guess we can provide support in Chinese. Maybe @xrain0610, @jethrokuan and me can help.

herrbischoff:
While support in Chinese sounds great on the surface, it will make it impossible for any non-Chinese speakers (like me) to benefit from the questions and answers. I guess you could argue the other way around, that non-English speakers cannot benefit from replies in all English. That assessment would be correct. Then again, English appears to be the agreed-upon *lingua franca* of the tech world. Today, it is virtually impossible to write code without at least a basic grasp of English language fundamentals. Whether that should remain so is another (no doubt very interesting and highly challenging) discussion in itself. For now, that's the state of affairs. It may seem counter-intuitive and to some degree exclusive. I would prefer to keep discussion and issues English-only, to give the broadest possible access to those interested in Fisherman.

Hopefully my comment is not ill-received. I'm not a native speaker myself but I still learned English out of necessity to code, understand tech and effectively communicate internationally.

jethrokuan:
@herrbischoff I echo your concerns, but I disagree that keeping it English-only

> (gives) the broadest possible access to those interested in Fisherman.

I'm willing to try and help Chinese-wise, but it's similar to doing math, or physics in Chinese. It's possible that I won't understand what they're saying either. That works the other way round as well. Chinese users may not be able to express their problems as well in english as they could in chinese, and we'd have problems trying to debug their problems.

What we could do in the meantime is to provide accurate translations of the problem and the solution. It may be extra effort, but I'm thinking the number of Chinese issues here would be small anyway.

What I'm saying is that we could give this a shot, and if it doesn't work/the direction this is going is wrong, I'm all for dropping Chinese support.

pickfire:
@herrbischoff We isn't here for dropping english support. In my humble opinion, I guess adding chinese support could benefit fisherman as although english is the international language, not everyone speak it well (a large part of the world speak chinese too, maybe comparable with english).

If we have the ability to provide support in chinese I guess it is nice to just do it. If they speak really bad in english then it will be no different than talking to them without understanding while wasting all the time.

I know that the majority of the contributers isn't chinese speakers, we can solve this by translating the summary of the problem into english so that non-chinese speakers can.

bucaran:
@xrain0610 Like this:

```md
Fisherman is a blazing fast, modern plugin manager for fish.

> Fisherman 现在提供中文支持,如果你有任何问题,可以以中文提交ISSUE,我们会尽快解决,感谢您的支持! 
```

jethrokuan:
> Fisherman 现在提供中文支持,若有疑问,可以中文提交ISSUE,我们会尽快解决,感谢您的支持!

herrbischoff:
> Chinese users may not be able to express their problems as well in english as they could in chinese, and we'd have problems trying to debug their problems.

@jethrokuan: I see what you mean. The same is probably true for the more than 6000 distinct languages spoken around the globe. I know that language shapes thought to a major degree and transferring that thought into a whole different structure (≈ language) may sometimes be challenging or even impossible.

On a side note, after thinking about it a little more, I noticed the following: to be completely honest, I was surprised to catch myself wanting to feel treated unfair. WTF? Because I had to adapt and now others don't have to. Jealousy, completely misplaced. Interesting reaction — and sadly all too human I suppose. 

I think that English summaries of the problem after solving are the way to go. Otherwise I'll try to unleash Google Translate on the issue. ;)

bucaran:
Thanks!

PR #206, opened 2016-04-21
Add missed files

sijad:


detailyang:
LGTM ping @bucaran

bucaran:
@sijad Can you just push the .gitignore for now. I am not sure we should keep using .editorconfig and the reason is that there is a `fish_indent` function that can help us format the code similar to what `go fmt` does for golang.

For this we would need to create a simple plugin to wrap the function, but it would be nice to have.

Please tell me if that has nothing to do with .editorconfig or if I missed something.

sijad:
done ;-)

sijad:
let me change the commit message

sijad:
good to go

PR #210, opened 2016-04-22
Fix link in readme

domoritz:


bucaran:
Thanks @domoritz, all README pages had the same dead link, so I fixed this myself in one swoop.

3e187c5

PR #231, opened 2016-04-26
eliminate -F for ln for OpenBSD by abstracting ln flags

bonds:
OpenBSD's ````ln```` does not support the -F flag. This PR drops the flag, only for OpenBSD, so as not to change the behavior for other OSes.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -39,6 +39,12 @@ function fisher
         set -g fisher_bundle "$fish_config/fishfile"
     end
 
+    if test (uname -s) = "OpenBSD"
+        set ln_flags = "-sf"
+    else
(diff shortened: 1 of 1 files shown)


bucaran:
LGTM. Thanks!

bucaran:
Good news from the `manpage`:

     -F    If the target file already exists and is a directory, then remove it so that the link may occur.  The
           -F option should be used with either -f or -i options.  If none is specified, -f is implied.  The -F
           option is a no-op unless -s option is specified.


In other words, it seems we can actually drop -F entirely.

@bonds Do you agree with my interpretation?

bucaran:
@bonds Nevermind, what it's saying is not what I meant.

Thanks again.

Issue #232, opened 2016-04-27
Update error


amio:
On osx 10.11.4
```
▸ Updating Fisherman
cat: /Users/Amio/.local/share/fisherman/VERSION: No such file or directory
Aye! Fisherman 15 updated (0s)
fish: Unknown command '__fisher_plugin_validate'
~/.local/share/fisherman/functions/fisher_update.fish (line 1): __fisher_plugin_validate $item
                                                                ^
in command substitution
	called on line 0 of file ~/.local/share/fisherman/functions/fisher_update.fish

in function 'fisher_update'
	called on line 1 of file -

from sourcing file -
	called on line 60 of file /usr/local/Cellar/fish/2.2.0/share/fish/functions/eval.fish

in function 'eval'
	called on line 51 of file ~/.local/share/fisherman/functions/fisher.fish

in function 'fisher'
	called on standard input
	with parameter list 'update'

fisher: '' is not a valid name, path or URL.
fish: Unknown command '__fisher_plugin_validate'
~/.local/share/fisherman/functions/fisher_update.fish (line 1): __fisher_plugin_validate $item
                                                                ^
in command substitution
	called on line 0 of file ~/.local/share/fisherman/functions/fisher_update.fish

in function 'fisher_update'
	called on line 1 of file -

from sourcing file -
	called on line 60 of file /usr/local/Cellar/fish/2.2.0/share/fish/functions/eval.fish

in function 'eval'
	called on line 51 of file ~/.local/share/fisherman/functions/fisher.fish

in function 'fisher'
	called on standard input
	with parameter list 'update'

fisher: '' is not a valid name, path or URL.
fish: Unknown command '__fisher_plugin_validate'
~/.local/share/fisherman/functions/fisher_update.fish (line 1): __fisher_plugin_validate $item
                                                                ^
in command substitution
	called on line 0 of file ~/.local/share/fisherman/functions/fisher_update.fish

in function 'fisher_update'
	called on line 1 of file -

from sourcing file -
	called on line 60 of file /usr/local/Cellar/fish/2.2.0/share/fish/functions/eval.fish

in function 'eval'
	called on line 51 of file ~/.local/share/fisherman/functions/fisher.fish

in function 'fisher'
	called on standard input
	with parameter list 'update'

fisher: '' is not a valid name, path or URL.
fish: Unknown command '__fisher_plugin_validate'
~/.local/share/fisherman/functions/fisher_update.fish (line 1): __fisher_plugin_validate $item
[...]

PR #233, opened 2016-04-27
Preserve meaningful prefix(plugin-, theme-, etc) for packages. #232

amio:


bucaran:
@amio Thank you so much! :) 

But I think there was a misunderstanding, we can't remove those checks, because there are plugins and themes in oh-my-fish that use those prefi already.

detailyang:
yep, it's historical reason :alien: @amio

amio:
Ooops, thanks for point out ;-)

PR #252, opened 2016-05-04
use /bin/date replace perl to get get_epoch_in_ms

detailyang:
in my virtual box
````bash
[root@localhost ~]# cat /etc/issue
CentOS release 6.5 (Final)
Kernel \r on an \m
````

and when i run command
````bash
fisher simple
````

i got 
````bash
Can't locate Time/HiRes.pm in @INC (@INC contains: /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .).
````

i find we use
 
````bash
perl -MTime::HiRes -e 'printf("%.0f\n", (Time::HiRes::time() * 1000) - '$elapsed')'
````
 to get millisecond, but in my host i dont have install Time-HiRes module.  i do install with `yum install perl-Time-HiRes` so that resolve it

are there something i missing?  so just use /bin/date to replace.

code LGTM. But i'm not sure whether it's always right. maybe we should add test:)

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -1740,7 +1740,7 @@ function __fisher_get_epoch_in_ms -a elapsed
         set elapsed 0
     end
 
-    perl -MTime::HiRes -e 'printf("%.0f\n", (Time::HiRes::time() * 1000) - '$elapsed')'
+    math (command date +%s%3N) - $elapsed
 end
 
 
--- a/fisher.fish
+++ b/fisher.fish
@@ -1740,7 +1740,12 @@ function __fisher_get_epoch_in_ms -a elapsed
         set elapsed 0
     end
 
-    math (command date +%s%3N) - $elapsed
+    switch (uname)
+        case Darwin
+            perl -MTime::HiRes -e 'printf("%.0f\n", (Time::HiRes::time() * 1000) - '$elapsed')'
+        case '*'
+            math (command date +%s%3N) - $elapsed
+    end
 end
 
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by detailyang** [comment]
ping @bucaran

**Comment by bucaran** [comment]
@detailyang That won't work on OS X, that does not expose the millisecond count since the epoch via date.

We need to check the OS and create Linux/OSX specific epoch functions.

**Comment by detailyang** [comment]
oops, so we should use `uname` to check os type

**Comment by detailyang** [comment]
how about this? @bucaran

**Comment by bucaran** [comment]
Yes, that's better. The best way to go is to create a function in fisher.fish.

```
if operating system is osx
   function __fisher_get_epoch
          # use perl here
   end   
else
   function __fisher_get_epoch
          # use linux
   end           
end
```

**Comment by detailyang** [comment]
wow, fish can define function like this!!!

**Comment by detailyang** [comment]
done

**Comment by bucaran** [comment]
@detailyang Looks good. Is it working good for you?

**Comment by detailyang** [comment]
yes, i tested on osx and linux . LGTM:)

**Comment by bucaran** [comment]
LGTM @Markcial @jethrokuan ?

**Comment by jethrokuan** [comment]
after a huge struggle with skylake linux support i ended up on ubuntu 16.04 LTS. code still works for me. LGTM.

**Comment by bucaran** [comment]
What does `%3` mean in

```fish
command date +%s%3N
```
?

bucaran:
The 3 truncates the nanoseconds so we get milliseconds, which is what we want.

* http://serverfault.com/questions/151109/how-do-i-get-current-unix-time-in-milliseconds-using-bash

Issue #312, opened 2016-06-09
github api rejects calls when `fisher ls-remote` is run too frequently


ianferguson:
Currently if `ls-remote` is run too frequently, fisher will begin returning the following error, 

```
! I could not update the remote index.

This is most likely a problem with http://api.github.com/
or a connection timeout. If the problem persists, open an
issue in: <github.com/fisherman/fisherman/issues>
```

This might potentially be due to github throttling/rejecting anonymous api searches occurring above a similar rate, similar to the conditions that led to homebrew to support passing an API token to homebrew for searches: https://gist.github.com/christopheranderton/8644743

Would it be possible to enable similar facilities in fisher that would allow users to provide an API token to fisherman for the github call behind ls-remote?

brj:
Hi @ianferguson.

* GITHUB_USER and 
* GITHUB_TOKEN 

are available.

```fish
set -x GITHUB_USER ianferguson
set -x GITHUB_TOKEN xxxxxx
```

ianferguson:
Awesome, thanks! Would it be helpful if I opened a PR to updates the docs
somewhere with that info?

brj:
@ianferguson Yes, it would be! Can you add a new question to the FAQ?

ianferguson:
opened #314

PR #314, opened 2016-06-10
Add FAQ for resolving github api token issues

ianferguson:
Add a FAQ section for how to resolve ls-remote being rate limited/rejected by the GitHub api when accessing anonymously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,3 +162,25 @@ A plugin is:
 ### How can I list plugins as dependencies to my plugin?
 
 Create a new fishfile at the root level of your project and write the plugin URL such as *github.com/owner/repo*.
+
+### Why am I receiving errors when running `fisher ls-remote`?
+
+If you are consistently seeing the following error while repeatedly running `fisher ls-remote`:
+
+```
+! I could not update the remote index.
+
+This is most likely a problem with http://api.github.com/
+or a connection timeout. If the problem persists, open an
+issue in: <github.com/fisherman/fisherman/issues>
+```
+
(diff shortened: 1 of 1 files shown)


brj:
LGTM!

ianferguson:
I'm on a different computer this morning and trying to reproduce the root issue (just to double check that the instructions I put work end to end) and am not actually able to get it to reproduce, even when I run `while true; fisher ls-remote; end` for a few minutes.

I'm going to look into this a little more over the weekend, but probably shouldn't merge this until its clear that the origin issue in #312 is _actually_ an issue, sorry!

brj:
Sounds reasonable. I'll assign you to this issue then. Merge when you are ready please :)

brj:
@ianferguson Okay, so here's the thing. I knew that many unauthenticated requests to the GitHub API would quickly reach the limit, so no matter how many times you run ls-remote, fisherman will only update the index once every 30 minutes IIRC. That's what may be happening here, thus you can't repro it again.

ianferguson:
@brj good flag! I should have time either a bit later today, or tomorrow to set up a repro that induces the GitHub api limit without using fisher ls-remote in that case. Probably overkill at this point, but just want to make sure I cross my t's and dot my i's before merging.

ianferguson:
Just confirmed that once GitHub's search api is rejecting anonymous requests, following the instructions in this PR will resolve the issue.

brj:
Thanks and welcome aboard :)

PR #351, opened 2016-11-19
Explain how to show a plugin's description

asymmetric:
Adds an example of how to show the description of a plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,12 @@ Show everything that's available.
 fisher ls-remote
 ```
 
+Show a plugin's description.
+
+```
+fisher ls-remote fzf
+```
+
 Update everything.
 
 ```


edouard-lopez:
related #352.

@jbucaran should we go for working examples (e.g. `fisher ls-remote fzf`) or man page-like (e.g. `fisher ls-remote [package]`.

Having copy-paste ready has some drawbacks, but non-working are annoying. What do you think?

jbucaran:
Thanks @asymmetric! 

@edouard-lopez Working examples are nicer, but no hard rules. I am unsure about documenting this bit, after all @asymmetric's other PR already illustrates this, kinda.

edouard-lopez:
@jbucaran yep add the same hesitation. Maybe listing packages on a one-per-line basis with their description will be better. Many packages manager do that (`apt`, `npm`, etc.) so you got enough information. Cause right now some package have very low information smell. For instance:

    omf/gi                          omf/theme-bira                  omf/theme-taktoa

jbucaran:
That's what ls-remote --format does. So, you are suggesting to present more information by default. That would make sense. I actually prefer just using fish pager, that fisherman already fills in with the relevant info, e.g

screen shot 2016-11-21 at 1 56 43

PR #352, opened 2016-11-19
Explain `ls-remote --format` option

asymmetric:
Add example of using `ls-remote --format` to display additional info about available plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,12 @@ Show everything that's available.
 fisher ls-remote
 ```
 
+Show additional information about available plugins:
+
+```
+fisher ls-remote --format="%name(%stars): %info [%url]\n"
+```
+
 Update everything.
 
 ```


jbucaran:
@asymmetric Thanks!!

edouard-lopez:
Thanks @asymmetric 
@jbucaran updated the French version.

I just noticed this by chance, is there a way to keep translation update? First would be to tag such issues/PR as `translation`-related.

jbucaran:
@edouard-lopez

>  Is there a way to keep translation update?

Tough one. We could ping the previous translators, but still.

BTW I just created a _translation_ tag. Please use it accordingly!

edouard-lopez:
@jbucaran I can't set labels, you [need `Write permissions` or `Admin permissions`](https://help.github.com/articles/repository-permission-levels-for-an-organization/) to do so.

jbucaran:
@edouard-lopez Fixed that for everyone!

PR #367, opened 2017-02-10
Improve warnings about fish versions

### LINKED ISSUE #362: Snippets 2.3.0+ compatibility
As an example, `conf.d/` *snippets* were introduced in `fish v2.3.0`.

If plugins start using *snippets*, but the user is still on `v2.2.0` - like a lot of Ubuntu users - the plugin will stop working silently.

Are there any solutions for fisherman to warn the user about incompatibilities like this, or does every plugin need to check the installed fish version and print a warning.

aristidesfl:
Check commits for details

Closes #362

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -28,6 +28,30 @@ function $fisher_cmd_name -d "fish plugin manager"
             end
 
             return 1
+
+        case 2.2.0
+            __fisher_log info "
+                You need fish 2.3.0 or higher to take advantage of snippets.
+                Without it some plugins might not work.
+            "
+
+            if type -q brew
+                __fisher_log info "Please run &brew upgrade fish&"
+            else
+                __fisher_log info "
+
+                    Refer to your package manager documentation
+                    for instructions on how to upgrade your fish build.
+
+                    If you can not upgrade, append the following code
+                    to your ~/.config/fish/config.fish:
+
+                    &for file in ~/.config/fish/conf.d/*.fish&
+                    	&source $file&
+                    &end&
+
+                "
+            end
     end
 
     set -g fisher_version "2.12.0"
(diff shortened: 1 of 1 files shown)


aristidesfl:
Shall we merge?

jbucaran:
👍

fisher.fish
```diff
@@ -16,18 +16,42 @@ end
 function $fisher_cmd_name -d "fish plugin manager"
     switch "$FISH_VERSION"
         case 2.1.2 2.1.1 2.1.0 2.0.0
-            __fisher_log error "You need fish &2.2.0& or higher to use fisherman."
+            echo "You need fish 2.2.0 or higher to use fisherman."
 
-            if command -s brew > /dev/null
-                __fisher_log info "Run &brew up; brew upgrade --HEAD fish&"
+            if type brew >/dev/null 2>&1
```
aristidesfl: `command -s` was only introduced in `fish 2.2.0` (a version we are sure the user doesn't have installed) . Without `-s` it actually executes the command even with `>/dev/null 2>&1`. `type` doesn't

fisher.fish
```diff
@@ -16,18 +16,42 @@ end
...
 
             return 1
+
+        case 2.2.0
+            __fisher_log info "
+                You need fish 2.3.0 or higher to take advantage of snippets.
+                Without it some plugins might not work.
+            "
+
+            if type -q brewe
```
aristidesfl: good catch ;)

fisher.fish
jbucaran: @aristidesfl You are totally right! 👍

jbucaran:
@aristidesfl :shipit:

Issue #364, opened 2017-02-10
Document how to install branches


aristidesfl:
- [ ] REAME.MD
- [ ] man pages / fisher help

PR #369, opened 2017-02-15
Document how to install branches and tags

aristidesfl:
Closes: #364

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,16 @@ Install more plugins.
 fisher fzf edc/bass omf/thefuck omf/theme-bobthefish
 ```
 
+Install a specific branch.
+```sh
+fisher edc/bass:master
+```
+
+Install a specific tag.
+```sh
+fisher edc/bass@1.2.0
+```
+
 Install a gist.
 
 ```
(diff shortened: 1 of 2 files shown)


jbucaran:
Nice gesture adding it to the man page too. Do you think it would sound more natural to say:

> Install by tag.
>Install by branch name.

I'm not sure.

aristidesfl:
I would say no, but up to you. @laughedelic ?

Related: https://medium.com/dropbox-design/design-words-with-data-fe3c525994e7#.k3uxwqkbo

jbucaran:
Nice link. Okay, let's go with what you wrote. @laughedelic's feedback would also be appreciated.

laughedelic:
I think it's better the way @aristidesfl wrote it 👌

- "... by ..." sounds for me like "by the means of ...", which is not applicable. You _refer_ to what you want to install by a tag/branch, but you install by... I don't know, running this command, using fisherman...
- "install from tag" just sounds strange to me. A tag is rather a fixed point in time, while here is feels like a location/place/source.

I'm not a native English speaker though 😅

jbucaran:
@aristidesfl lgtm

jbucaran:
Thank you @laughedelic.

PR #383, opened 2017-05-15
Termux compatibility

christianrondeau:
This solves a few issues that prevents fisher from workin in [termux](https://termux.com/):

* Use `>&2` to redirect to stderr (this is more portable than `/dev/stderr`): https://unix.stackexchange.com/a/274044/68749
* Use `/proc/self/fd/2` for `__fisher_stderr` (only if the OS is termux, to minimize risk)
* Use `$PREFIX` to check which OS is running (if someone has a better idea, I'm sure there's a better way)
* Use `mandoc` directly for `fisher help` on termux, because it runs the [busybox](https://busybox.net/about.html) version of `man` which does not support `-P`

I have tested with fish `2.5.0` on:

* `Termux`
* `Mint` (a recent version)
* `Ubuntu 16.04.2`

I tested `fisher`, `fisher z`, `fisher help` and `fisher update`.

I think another pair of eyes would be warranted since I have limited experience with `fish` and `linux` in general.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -56,15 +56,31 @@ function $fisher_cmd_name -d "fish plugin manager"
 
     set -g fisher_version "2.13.0"
     set -g fisher_spinners ⠋ ⠙ ⠹ ⠸ ⠼ ⠴ ⠦ ⠧ ⠇ ⠏
-    set -g __fisher_stdout /dev/stdout
-    set -g __fisher_stderr /dev/stderr
+
+    if [ -e /dev/stdout ]
+      set -g __fisher_stdout /dev/stdout
+    else if [ -e /proc/self/fd/1 ]
+      set -g __fisher_stdout /proc/self/fd/1
+    else
+      echo "No suitable stdout"
+      return 1
+    end
+
+    if [ -e /dev/stderr ]
+      set -g __fisher_stderr /dev/stderr
+    else if [ -e /proc/self/fd/2 ]
+      set -g __fisher_stderr /proc/self/fd/2
+    else
+      echo "No suitable stderr"
+      return 1
+    end
 
     function __fisher_show_spinner
         if not set -q __fisher_fg_spinner[1]
             set -g __fisher_fg_spinner $fisher_spinners
         end
 
-        printf "  $__fisher_fg_spinner[1]\r" > /dev/stderr
+        printf "  $__fisher_fg_spinner[1]\r" >&2
 
         set -e __fisher_fg_spinner[1]
     end
@@ -110,7 +126,7 @@ function $fisher_cmd_name -d "fish plugin manager"
             return
 
         case -h
-            __fisher_usage > /dev/stderr
+            __fisher_usage >&2
             return
     end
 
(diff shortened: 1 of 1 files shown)


aristidesfl:
Will test soon

fisher.fish
```diff
@@ -56,15 +56,21 @@ function $fisher_cmd_name -d "fish plugin manager"
 
     set -g fisher_version "2.12.0"
     set -g fisher_spinners ⠋ ⠙ ⠹ ⠸ ⠼ ⠴ ⠦ ⠧ ⠇ ⠏
-    set -g __fisher_stdout /dev/stdout
-    set -g __fisher_stderr /dev/stderr
+
+    if [ "$PREFIX" = "/data/data/com.termux/files/usr" ]
```
jbucaran: Do we need to hardcode this stuff?

christianrondeau: I could instead check if the file exists (`[ - f /dev/stderr ] `), that would be cleaner, but I'm not sure how to detect the man version. I thought this is simple, pretty safe and has a minimum risk on other platforms but I agree checking the file existence would be better. I'll update the PR. 

I'll comment where this condition would be "more problematic" separately.

christianrondeau:
I have made the change for__fisher_stderr and__fisher_stdout. Tested in Termux and Mint. 

I also fixed a remaining /dev/stderr in__fisher_log

Only the mandoc thing remains!

jbucaran:
@fisherman/all 

If anyone is using termux, can you please checkout @christianrondeau's branch and test this out? 🙏

christianrondeau:
Might be obvious, but you can either clone the repo, switch to the termux branch and us `ln -s ~/fisherman/fisher.fish ~/.config/fish/functions/fisher.fish` and `. ~/.config/fish/functions/fisher.fish`, OR; you can download the fisher.fish file directly from my branch, like instructed in the README.

Hopefully, I'm not alone wanting to use Termux :)

christianrondeau:
Hey! Just rebased this branch, any chance for it to get merged?

jbucaran:
@christianrondeau If you've tested it and it works and no one else opposes this change, I don't have a problem with it getting merged.

jethrokuan:
Doesn't break on other terms, such as in urxvt, so it's fine with me

jbucaran:
@christianrondeau What about people _not_ using termux or using fisherman on windows?

Also, changes should be followed by a bump in the version number hardcoded in the source code. 👋

christianrondeau:
Initially I tested on Ubuntu 16 and Mint, I use it on Termux and on lxss (i.e. Windows, but I stopped using cygwin a while ago), and I just verified and it also works on Alpine Linux :)

About the version number, you're asking me to bump it in my PR? Should I bump it to `2.13.1` since there is no API change?

Let me know if there's something more expected of me, so that I can stop using my custom branch :)

jbucaran:
@christianrondeau Let's merge, then let's bump. 

> Should I bump it to 2.13.1 since there is no API change?

Sounds reasonable. Let's do that after merging here.

jbucaran:
```
fisher update
```

Everything seems to be running okay! :)

- https://github.com/fisherman/fisherman/commit/3e74f7bd4acd2d8441c2b5e6ad60d075a7d19989

christianrondeau:
Pushed the version bump. Thanks!

christianrondeau:
(oops but since it's merged, I'd need to create a new PR, should I?)

jbucaran:
No worries, it's done: 3e74f7b 👋

PR #384, opened 2017-05-21
Add https:// to install script

aristidesfl:
Should we add https:// to the installation instruction for security purposes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Translations: [日本語], [繁體中文], [简体中文], [한국어], [Рус�
 ## Install
 
 ```sh
-curl -Lo ~/.config/fish/functions/fisher.fish --create-dirs git.io/fisher
+curl -Lo ~/.config/fish/functions/fisher.fish --create-dirs https://git.io/fisher
 ```
 
 ## Usage


jbucaran:
git.io is a domain managed exclusively by GitHub. I can't really change where the link points to, unless I GitHub to remove it. Then I could create a new redirect. I think that has nothing to do with using https or http. 

What does `curl` use by default if none is given? (such as is the case here).

jbucaran:
> https://curl.haxx.se/docs/manpage.html

> If you specify URL without protocol:// prefix, curl will attempt to guess what protocol you might want. It will then default to HTTP but try other protocols based on often-used host name prefixes. For example, for host names starting with "ftp." curl will assume you want to speak FTP.

aristidesfl:
It defaults to http. `http://git.io` redirects to `https://git.io` but its already too late.

jbucaran:
@aristidesfl What's already too late? 😄

aristidesfl:
Means that we should use https use merge this, because the redirect is worthless.

jbucaran:
@aristidesfl Sorry, can you say that again? 🤔

liamogdene:
ok
contact me here:: newsex996@gmail.com

jbucaran:
git.io/fisher redirects to https://git.io/fisher.

liamogdene:
ok
contact me here:: newsex996@gmail.com

aristidesfl:
The initial connection is insecure, so redirections to secure urls don't solve the problem, since the connection can be hijacked right from the begging, without the user suspecting of anything.

jbucaran:
@aristidesfl In order to hijack the url, you'd need to hijack github's git.io vanity service in the first place. If someone is able to hack the url to redirect somewhere else, it won't matter whether it is http or https.

aristidesfl:
@jbucaran I'm talking about MITM.

jbucaran:
@aristidesfl Can you hack it?

aristidesfl:
Sure, anyone who has the ability to intercept you traffic can execute arbitrary code in your machine. That includes random individuals in your local network. Outside of that, ISP's, state adversaries and other entities with access to the infrastructure.

albinekb:
It should be `https`!

jbucaran:
https://github.com/fisherman/fisherman/commit/152c9d2089fde381fbf7d66ab0cba1a3c509cfd6

jbucaran:
🎉 and thanks to @aristidesfl.

albinekb:
Awesome @jbucaran really appreciate it! ❤️

PR #387, opened 2017-06-09
Preserve symlinked fish_user_key_bindings

### LINKED ISSUE #343: Symlinked key bindings
Installing new plugin with key bindings breaks symlinked `~/.config/fish/functions/fish_user_key_bindings.fish`, i.e. it creates a new file in place of symlink. Is this expected behavior, and if so, where should I store my own version controlled key bindings?

Perlence:
Resolves #343. Removing plugin won't remove empty
fish_user_key_bindings.fish if it's a symlink.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -1492,8 +1492,7 @@ function __fisher_key_bindings_remove -a plugin_name
 
     fish_indent < "$user_key_bindings" | command sed -n "/### $plugin_name ###/,/### $plugin_name ###/{s/^ *bind /bind -e /p;};" | builtin source ^ /dev/null
 
-    command sed "/### $plugin_name ###/,/### $plugin_name ###/d" < "$user_key_bindings" > "$user_key_bindings.$tmp"
-    command mv -f "$user_key_bindings.$tmp" "$user_key_bindings"
+    command sed --follow-symlinks -i "/### $plugin_name ###/,/### $plugin_name ###/d" "$user_key_bindings"
 
     if command awk '
         /^$/ { next }
@@ -1511,7 +1510,7 @@ function __fisher_key_bindings_remove -a plugin_name
             exit 1
         }
 
-    ' < "$user_key_bindings"
+    ' < "$user_key_bindings" -a ! -L "$user_key_bindings"
 
         command rm -f "$user_key_bindings"
     end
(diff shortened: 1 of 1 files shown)


fisher.fish
```diff
@@ -1492,8 +1492,7 @@ function __fisher_key_bindings_remove -a plugin_name
 
     fish_indent < "$user_key_bindings" | command sed -n "/### $plugin_name ###/,/### $plugin_name ###/{s/^ *bind /bind -e /p;};" | builtin source ^ /dev/null
 
-    command sed "/### $plugin_name ###/,/### $plugin_name ###/d" < "$user_key_bindings" > "$user_key_bindings.$tmp"
-    command mv -f "$user_key_bindings.$tmp" "$user_key_bindings"
+    command sed --follow-symlinks -i "/### $plugin_name ###/,/### $plugin_name ###/d" "$user_key_bindings"
```
jbucaran: I am afraid `-i` could break in a platform that does not support it. Can we use the same code as before, and only add `--follow-symlinks`?

jbucaran: @Perlence Left you a comment ↑.

Perlence: I guess we can, but that `mv` should be replaced with `cat > ...; and rm` as seen in the next chunk.

jbucaran: @Perlence Hmm, and isn't there a way to force `mv` to work like we want instead? 🤔

Perlence: @jbucaran Not that I can think of. A similar problem is tackled in Vim with `backupcopy` option: http://vimdoc.sourceforge.net/htmldoc/options.html#%27backupcopy%27

jbucaran: @Perlence Okay, then, let's go with the alternative that does not break fisherman! 😄

Perlence: @jbucaran That should do it.

jbucaran: 👀

e-beach:
Is there a reason this isn't merged?

jbucaran:
@e-beach We could merge it, but need more input from users that it's working as expected.

Does it work for you?

e-beach:
I tried uninstalling and installing that function, and it works for me.

jbucaran:
Hi @Perlence! It's been a while. Have you been using fisherman lately with your modifications and noticed anything wrong?

amir20:
I am going to close this since it hasn't gotten any responses for over a year now.

PR #392, opened 2017-06-29
Fix #391

### LINKED ISSUE #391: fisher.fish completion always installed into wrong folder
I keep my fisherman install in a separate folder from my user config. In my `config.fish` I have:

```
set fish_path (dirname (status -f))/fisherman
set fish_function_path $fish_path/functions $fish_function_path
set fish_complete_path $fish_path/completions $fish_complete_path
for file in $fish_path/conf.d/*.fish
    source $file
end
```

Everything is installed into the correct location except for the fisher.fish completion. It's installed to `~/.config/fish/completions/fisher.fish`, instead of `~/.config/fish/fisherman/completions/fisher.fish` like everything else.

externl:
Fisherman was always writing its completion file to $fish_config/completions.
This was incorrect behaviour if $fish_path was set to a directory other than
$fish_config (its default).

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -117,7 +117,7 @@ function $fisher_cmd_name -d "fish plugin manager"
     command mkdir -p {"$fish_path","$fish_config"}/{conf.d,functions,completions} "$fisher_config" "$fisher_cache"
     or return 1
 
-    set -l completions "$fish_config/completions/$fisher_cmd_name.fish"
+    set -l completions "$fish_path/completions/$fisher_cmd_name.fish"
 
     if test ! -e "$completions"
         echo "$fisher_cmd_name --complete" > "$completions"


fisher.fish
```diff
@@ -117,7 +117,7 @@ function $fisher_cmd_name -d "fish plugin manager"
     command mkdir -p {"$fish_path","$fish_config"}/{conf.d,functions,completions} "$fisher_config" "$fisher_cache"
     or return 1
 
-    set -l completions "$fish_config/completions/$fisher_cmd_name.fish"
+    set -l completions "$fish_path/completions/$fisher_cmd_name.fish"
```
jbucaran: If you put fisher completions in a different directory that is **not** in $fish_complete_path, then the completions will not be picked up.

jbucaran: @externl

externl: With `$fish_path` set currently, all completions get installed to `$fish_path/completions/` anyway, this just makes `$fisher_cmd_name.fish` follow the same behaviour.

jbucaran: @externl No, that's different. Fisherman will correctly take care of plugins it install and make sure functions/completions/snippets are available after they are installed regardless of their location.

The thing is, fisher completions should be in $fish_complete_path, otherwise `fisher` completions won't work.

externl: But, if fisher.fish is itself in `$fish_path/conf.d`, all files it generates should be in `$fish_path` as well. By that logic `$fish_path` should be ignored when installing the `fisher.fish` function. It should be all or none.

externl: @jbucaran 

> Fisherman will correctly take care of plugins it installs and make sure functions/completions/snippets are available after they are installed regardless of their location.

Unless I'm mistaken this will only work on initial install if `$fish_path` is set and the user does not manually update `fish_function_path` and `fish_function_path `. The next time they load fish these things will not be available.

jbucaran: @externl Where did you install fisherman when you downloaded it? :thi

externl: Hmm... originally it was in `$fish_config/functions`, but now it's in `$fish_path/functions`.

jbucaran: @externl It isn't, but you are responsible for setting up fish_path and adding dirs to your $fish_{function/complete}_path yourself as well as loading snippets on startup (which I see you are doing).

Hmm 🤔 🙃

I think this might be solid.

jbucaran:
Reviewed and merged! 💥👏🎉 Thanks, @externl!

PR #401, opened 2017-09-18
Allow for copy instead of linking

### LINKED ISSUE #400: Allow for copy instead of linking
I check my dotfiles into a git repo that I use across a few computers. They are also synced across some of the computers using Dropbox. The symlinks are causing issues, as they include the full path, and the home path on the systems I use are not all the same.

Please allow for the optional copy of plugin files instead of linking. I have a PR that does this, but felt like I should open the discussion first. Thanks!

skisulli:
I check my dotfiles into a git repo that I use across a few computers. They are also synced across some of the computers using Dropbox. The symlinks are causing issues, as they include the full path, and the home path on the systems I use are not all the same.

Please allow for the optional copy of plugin files instead of linking.

Fixes #400

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -116,6 +116,10 @@ function $fisher_cmd_name -d "fish plugin manager"
         set -g fisher_file "$fish_path/fishfile"
     end
 
+    if test -z "$fisher_copy"
+        set -g fisher_copy false
+    end
+
     switch "$argv[1]"
         case --complete
             __fisher_complete
@@ -800,7 +804,11 @@ function __fisher_plugin_enable -a path
             command mv -f "$target" "$backup_target" ^ /dev/null
         end
 
-        command ln -sf "$file" "$target"
+        if eval $fisher_copy
+            command cp -Rf "$file" "$target"
+        else
+            command ln -sf "$file" "$target"
+        end
 
         builtin source "$target" ^ /dev/null
 
@@ -815,19 +823,31 @@ function __fisher_plugin_enable -a path
 
     for file in $path/{functions/,}*.{py,awk}
         set -l base (basename "$file")
-        command ln -sf "$file" "$fish_path/functions/$base"
+        if eval $fisher_copy
+            command cp -Rf "$file" "$fish_path/functions/$base"
+        else
+            command ln -sf "$file" "$fish_path/functions/$base"
+        end
     end
 
     for file in $path/conf.d/*.{py,awk}
         set -l base (basename "$file")
-        command ln -sf "$file" "$fish_path/conf.d/$base"
+        if eval $fisher_copy
+            command cp -Rf "$file" "$fish_path/conf.d/$base"
+        else
+            command ln -sf "$file" "$fish_path/conf.d/$base"
+        end
     end
 
     for file in $path/conf.d/*.fish
         set -l base (basename "$file")
         set -l target "$fish_path/conf.d/$base"
 
-        command ln -sf "$file" "$target"
+        if eval $fisher_copy
+            command cp -Rf "$file" "$target"
+        else
+            command ln -sf "$file" "$target"
+        end
         builtin source "$target" ^ /dev/null
     end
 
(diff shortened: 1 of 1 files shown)


fisher.fish
```diff
@@ -815,27 +823,43 @@ function __fisher_plugin_enable -a path
 
     for file in $path/{functions/,}*.{py,awk}
         set -l base (basename "$file")
-        command ln -sf "$file" "$fish_path/functions/$base"
+        if eval $fisher_copy
```
JorgeBucaran: @skisulli `eval`?

JorgeBucaran:
@skisulli I think it creates links by default because of performance, but I can't remember anymore. 😄 

> The symlinks are causing issues, as they include the full path, and the home path on the systems I use are not all the same.

Why are the symlinks causing issues exactly?

Question: Is there any way we can accomplish this without introducing a new configuration variable?

fisher.fish
rsullivan-gpsw: I'm new to fish and then I had just `if $fisher_copy` the shell errored with the following message.

```
~/.config/fish/functions/fisher.fish (line 35): Variables may not be used as commands. In fish, please define a function or use 'eval $fisher_copy'.
        if $fisher_copy
           ^
in function '__fisher_plugin_enable'
	called on line 449 of file ~/.config/fish/functions/fisher.fish
	with parameter list '/Users/rsullivan/.config/fisherman/nvm'
```

What would be the proper way to check the value of a boolean?

rsullivan-gpsw: I'm new to fish and then I had just `if $fisher_copy` the shell errored with the following message.

```
~/.config/fish/functions/fisher.fish (line 35): Variables may not be used as commands. In fish, please define a function or use 'eval $fisher_copy'.
        if $fisher_copy
           ^
in function '__fisher_plugin_enable'
	called on line 449 of file ~/.config/fish/functions/fisher.fish
	with parameter list '/Users/rsullivan/.config/fisherman/nvm'
```

What would be the proper way to check the value of a boolean?

skisulli: I'm new to fish and then I had just `if $fisher_copy` the shell errored with the following message.

```
~/.config/fish/functions/fisher.fish (line 35): Variables may not be used as commands. In fish, please define a function or use 'eval $fisher_copy'.
        if $fisher_copy
           ^
in function '__fisher_plugin_enable'
	called on line 449 of file ~/.config/fish/functions/fisher.fish
	with parameter list '/Users/rsullivan/.config/fisherman/nvm'
```

What would be the proper way to check the value of a boolean?

JorgeBucaran: ```if test $fisher_copy = true```
or
```if test $fisher_copy = false```

**Comment by skisulli** [comment]
The full path varies between machines. So when I'm using Dropbox or git to sync the symlinks, they don't point to the file on each machine. For instance, my work computer Mac `$HOME` is `/Users/rsullivan`, my personal Mac is `/Users/ryan` and I have a linux machine that is `/home/ryan`.

The two Macs are synced with Dropbox, and when they sync they don't link to the correct place, because the link is the full `/Users/rsullivan/.config/fisherman/...` path.

The linux machine uses git to do the syncing. If I commit the symlink it also has the full path when pulled and therefore doesn't point to a file that exists on that machine.

File: fisher.fish (hunk 1 above)
Comment (skisulli) [reply fisher.fish]: removed the eval calls

**Comment by JorgeBucaran** [comment]
Thanks @skisulli! 

Can you also please update the docs by adding a new section to the [FAQ](https://github.com/fisherman/fisherman#faq)?

Similar to how `$fish_path` is documented [here].

**Comment by skisulli** [comment]
Will update docs. I was unable to think of a clean way to do this without a new config. If you have any thoughts on how to do it, I'm happy to try to implement.

**Comment by skisulli** [comment]
Updated the docs

**Comment by JorgeBucaran** [comment]
@skisulli Thanks!

File: README.md [hunk 2]
```diff
@@ -145,6 +145,16 @@ To customize this location:
 set -U fish_path ~/my/path
 ```
 
+### How do I have fisherman copy plugin files instead of linking
```
JorgeBucaran: Missed a `?`. 😉

JorgeBucaran:
@skisulli Perfect. Now, I'd like this to seat for a few days and see if anyone else has any feedback to share, then we'll merge! 

Thanks! 👋

skisulli:
Sounds good!

tpict:
Thanks so much for this @sullimander!

erikphansen:
Thank you for this! I found myself in the exact same position as @sullimander and am so happy to see there's now a way around this.

One other possible approach to [avoid using a global settings flag](
https://github.com/fisherman/fisherman/pull/401#issuecomment-330281536), and perhaps there are reasons this wouldn't work, is to pass an option flag when doing the install.

PR #407, opened 2017-12-12
`fisher up` obeys $fish_path

greenie:
Previously, running `fisher up` would always write fisher.fish to $fish_config/completions, even if $fish_path had been set.

This branch fixes that, and updates the README to reflect the extra steps required to source snippets and discover functions and completions installed by fisherman when using $fish_path.

Refs #391

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,10 +139,17 @@ The configuration and cache are saved to ~/.config/fisherman and ~/.cache/fisher
 
 The fishfile and plugins are saved to ~/.config/fish by default.
 
-To customize this location:
+To customize this location, add the following to your ~/.config/fish/config.fish file:
 
 ```fish
-set -U fish_path ~/my/path
+set -g fish_path ~/my/path
+
+for file in $fish_path/conf.d/*.fish
+  builtin source $file ^ /dev/null
+end
+
+set fish_function_path $fish_path/functions $fish_function_path
+set fish_complete_path $fish_path/completions $fish_complete_path
 ```
 
 ### What is a fishfile and how do I use it?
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -139,10 +139,17 @@ The configuration and cache are saved to ~/.config/fisherman and ~/.cache/fisher
 
 The fishfile and plugins are saved to ~/.config/fish by default.
 
-To customize this location:
+To customize this location, add the following to your ~/.config/fish/config.fish file:
 
 ```fish
-set -U fish_path ~/my/path
+set -g fish_path ~/my/path
```
JorgeBucaran: @greenie Why change the scope from -U to -g?

greenie: Ahh, looks like I'd copy-and-pasted the whole thing from my own fish config. I have -g instead of -U there as I felt that having fish_path written to ~/.config/fish/fishd.xxx every time I started fish seemed excessive. Happy to change it back to -U if I'm mistaken and that's a better thing to do.

greenie: True! I hadn't considered that. Changed it back to -U :)

JorgeBucaran:
@greenie LGTM! You already using this?

greenie:
@JorgeBucaran Yep, I've been giving it a go and running `fisher up` when I have `fish_path` set doesn't create the fisherman completions file in ~/.config/fish/completions any more. I've also run through the scenario of reinstalling fish and fisherman (from this branch) from scratch, and got the same behaviour 👍

JorgeBucaran:
@greenie Sounds good! Thank you for fixing this! 🎉

PR #412, opened 2018-03-21
improve index file parsing

sanyer:
For some reason index file was generated without space after first semicolon for me.
Using awk's match() function and RLENGTH to extract substring (either with space of without) resolves https://github.com/fisherman/fisherman/issues/399.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -1166,15 +1166,15 @@ function __fisher_remote_index_update
         }
 
         {
-            if (match($0, /^name:[[:space:]]?/)) {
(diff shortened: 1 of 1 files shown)


jorgebucaran:
@sanyer Maybe make the regex this:

```awk
/^name: */
```

instead of: 

```awk
/^name:[[:space:]]?/
```

sanyer:
`[[:space:]]` includes all whitespaces and also looks better. But `*` is indeed better than `?`.
Could probably match only horizontal whitespaces with `[[:blank:]]*`

jorgebucaran:
@sanyer 👍

PR #416, opened 2018-04-14
Use standard sh-style redirection for stderr

neersighted:
`^` has been removed as a stderr redirection operator (fish-shell/fish-shell#4394) in future versions of fish.


neersighted:
Ready to merge!

gyakovlev:
I squashed and pushed with slightly modified commit message.
Thanks!

PR #419, opened 2018-04-17
Fix awk parser to set correct variable. (ls-remote was failing to display %info tag correctly)

rholak:
The awk json parsing routine was setting a variable of "description"
instead of "info". This was causing a problem where the cached index -
and subsequently ls-remote commands - to show the "%info" format item as
the URL again.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -1171,7 +1171,7 @@ function __fisher_remote_index_update
             }
 
             if (match($0, /^description:[[:blank:]]*/)) {
-                description = substr($0, RLENGTH+1)
+                info = substr($0, RLENGTH+1)
             }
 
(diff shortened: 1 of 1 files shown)


gyakovlev:
testcase:

```shell-session
~> fisher ls-remote metro
github.com/fisherman/metro
github.com/fisherman/metro
```

after this fix:
```shell-session
~> rm ~/.cache/fisherman/.index 
~> fisher ls-remote metro
Powerline prompt optimized for speed and space
github.com/fisherman/metro
```

LGTM, thanks for the fix,
just need to re-fetch the cache for it to work right away.

Markcial:
@gyakovlev are you not able to merge? Or waiting for 2nd review?

PR #433, opened 2018-07-05
Fix config.fish example for custom paths

drrlvn:
The `fish_function_path` must be set before sourcing the `conf.d` files because they might call functions in that path.


jorgebucaran:
@drrlvn LGTM! Thanks! :)

PR #440, opened 2018-08-21
Update help to be more in line with README

erikphansen:
- Explain how to copy, not link, files when installing plugins
- This feature was added in v2.13.3


jorgebucaran:
Thank you, @erikphansen! 👍

PR #467, opened 2018-10-09
Store a list of all linked files under `installed.txt`

norcalli:
Create a list of installed files that are linked from package installation under `$fisher_config/installed.txt`.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -306,6 +306,7 @@ function _fisher_pkg_install -a pkg
                 end
         end
         echo "linking $target" | command sed "s|$HOME|~|" >&2
+        echo $target >> $fisher_config/installed.txt
         command ln -f $source $target
         switch $target
             case \*.fish


jorgebucaran:
@norcalli Seems rather arbitrary. Why do you need this?

norcalli:
I use it to check against a list of files to ignore from my committed dotfiles in my .gitignore, so that I can differentiate my defined files from the installed ones. I'm fine with maintaining it in a branch, but figured I might as well propose it.

jorgebucaran:
@norcalli Could you elaborate more? I still don't understand why you need it or why you couldn't accomplish what you are after without it.

If possible, I'd prefer if fisher gave you everything you need so that you (anyone) don't need to maintain a fork, but first, I need to understand the use case.

norcalli:
From my personal repository of configuration files, I link from `dotfiles/fish/` to `~/.config/fish`. Therefore, everything linked to `~/.config/fish/` is a part of my git repository. As a result, sometimes external tools create files in my folder, and I don't want to track those since they are 3rd party dependencies. Therefore, I manually created a list of all the files that I wanted to ignore before.

However, as these things update, those .gitignore entries become stale. Therefore, I wanted to automatically create the .gitignore entries. So I created a marker point `# fisher ignore` where I would put autogenerated entries. Then I made a script that runs an update, in the following two commands:

```
  sed -i -n '1,/^# fisher ignore/p' $DOTFILES/.gitignore
  cat ~/.config/fisher/installed.txt | sed 's:^.*/\.config/fish:dotfiles/fish:' | sort -u >> $DOTFILES/.gitignore
```

I attempted to create a list of files that would be relevant by scraping the output of `fisher`, but `fisher | grep '^linking'` didn't work, probably due to some kind of redirection to stderr, but no matter how many flags like `2>&1` or `2>/tmp/installed.txt`, I couldn't get it to store the output, so feeding the output of `fisher` was out of the question.

As a result I added that single line to store a list of the relevant files in a place I could use for a script.

jorgebucaran:
@norcalli Would [jorgebucaran/fisher#changing-the-installation-prefix](https://github.com/jorgebucaran/fisher#changing-the-installation-prefix) help?

jorgebucaran:
> @norcalli I couldn't get it to store the output, so feeding the output of fisher was out of the question. 

You couldn't parse the output because it was produced in a subshell.

FabioAntunes:
@norcalli my 2 cents, why do you symlink the entire `~/.config/fish` instead of symlinking folders why don't you just symlink the files?
That way there's need for maintaining a whitelist of files.

norcalli:
> @norcalli Wild idea. What about using the contents in the fishfile to generate the targets instead?
> 
> I'm open to other alternatives as I'd prefer if you (or anyone) didn't have to fork fisher so that it works for them.

The process of figuring out what the targets would be based on the fishfile would mean replicating the logic inside of `fisher` to resolve the names from the source to the output, considering that logic is not as simple as "copy it exactly," such as in the case of `*.fish` files in the root of the cloned git repository or prepending the name for `init.fish` and `key_bindings.fish`.

Now that I think about this more, it's not just a useful feature for that one facet, but for the purpose of uninstallation, you are duplicating the logic of resolving the source to target code in the `uninstall` method, whereas you could also just reference a list of installed files to use for the deletion part, which is arguably more dependable.

In the end, this solved a problem I need. So if your disposition is not to merge, then that's really quite ok. I already maintain my own repository of private forks with patches for other tools, so one more is not a problem.

@FabioAntunes Symlinking the whole directory means that when I do `funced` and `funcsave`, the files are immediately reflected in my `dotfiles` git repository, which is much more convenient, for example. In fact, at first I did symlink files individually, and found it to be more of an annoyance, and this approach is my evolution of that initial approach.

jorgebucaran:
@norcalli Are you aware of [jorgebucaran/fisher#changing-the-installation-prefix](https://github.com/jorgebucaran/fisher#changing-the-installation-prefix)?

norcalli:
> @norcalli Are you aware of [jorgebucaran/fisher#changing-the-installation-prefix](https://github.com/jorgebucaran/fisher#changing-the-installation-prefix)?

Yup. I think it would be more work to adapt that than the solution I settled upon. 
```
❯ git diff HEAD~1 dotfiles/fish/functions/fisher.fish
diff --git a/dotfiles/fish/functions/fisher.fish b/dotfiles/fish/functions/fisher.fish
index 07e9a09..88d975e 100644
--- a/dotfiles/fish/functions/fisher.fish
+++ b/dotfiles/fish/functions/fisher.fish
@@ -112,6 +112,19 @@ function _fisher_self_update -a file
     echo "fetching $url" >&2

     curl -s "$url?nocache" >$file@
+    echo 'diff --git a/dotfiles/fish/functions/fisher.fish b/dotfiles/fish/functions/fisher.fish
+index 24c43d7..07e9a09 100644
+--- a/dotfiles/fish/functions/fisher.fish
++++ b/dotfiles/fish/functions/fisher.fish
+@@ -306,6 +306,7 @@ function _fisher_pkg_install -a pkg
+                 end
+         end
+         echo "linking $target" | command sed "s|$HOME|~|" >&2
++        echo "$target" >> $fisher_config/installed.txt
+         command ln -f $source $target
+         switch $target
+             case \*.fish
+' | patch -p0 $file@

     set -l next_version (awk 'NR == 1 { print $4; exit }' < $file@)
     switch "$next_version"
```

I just patch it on update, haha, so I don't even need to fork it.

jorgebucaran:
@norcalli Still, I feel that we are missing an opportunity here. The resolution can't be to fork and patch fisher. I understand it works for you, and that's great. But we've already failed to help a future someone with a similar use case.

norcalli:
@jorgebucaran If someone else with a similar need arises, they can chose to bring it up again hopefully, or perhaps adapt a variation of my solution if they see it as sufficient. I wanted to gauge the opinion on the issue with the PR and to see if a better solution would be proposed. If it only solves my use case, it is not necessarily enough to merit merging in.

Therefore, when that future person arrives at this issue or creates a new issue, we can approach it with more certainty.

Issue #487, opened 2018-11-02
Custom install prefix guidance results in no custom prompt or default function overrides


cormacrelf:
The suggested code for a separate fisher install has:

```fish
set -g fish_function_path $fish_function_path $fisher_path/functions
set -g fish_complete_path $fish_complete_path $fisher_path/completions
```

Because `prefix/functions` is appended to `fish_function_path`, the system default function directories prevail.

```
$ type fish_prompt
fish_prompt is a function with definition
# Defined in /usr/local/Cellar/fish/2.7.1/share/fish/functions/fish_prompt.fish @ line 5
...
```

A better version, that lets the fisher functions override, lets you override them with your own in ~/.config/fish/functions, and does the same for completions:

```fish
# separate fisher install
set -l normal_fish $HOME/.config/fish
set -g fisher_path $HOME/.config/fish/erman
mkdir -p $fisher_path
set -g fish_function_path $normal_fish/functions $fisher_path/functions $fish_function_path 
set -g fish_complete_path $normal_fish/completions $fisher_path/completions $fish_complete_path
function source_confd
  for file in $fisher_path/conf.d/*.fish
    builtin source $file 2> /dev/null
  end
end
# fisherman bootstrap
if not test -f $fisher_path/functions/fisher.fish
  echo "... installing fisherman to $fisher_path ..."
  curl https://git.io/fisher --create-dirs -sLo $fisher_path/functions/fisher.fish
  fisher
end
source_confd
```

This seems like a lot of code for a pretty simple requirement -- is there any way to build some of it into the tool?

jorgebucaran:
> @cormacrelf ...is there any way to build some of it into the tool?

No, I don't think so, but I'm always open to suggestions.

cormacrelf:
Ok, but still probably allow ~/.config/fish/functions to beat the fisher ones. It's pretty useful for customizing a theme without forking the associated repo and changing your fishfile. Like `fish_mode_prompt`, which is fairly clunky in quite a lot of themes.

jorgebucaran:
@cormacrelf I agree we should prepend.

```fish
set fish_function_path $fisher_path/functions $fish_function_path
set fish_complete_path $fisher_path/completions $fish_complete_path
```

Can you send me a PR with that?

cormacrelf:
That's not what I mean. I mean

```
set -g fisher_path /path/to/other
set -l my_fish ~/.config/fish
[...]

PR #488, opened 2018-11-02
docs: fix function-loading precedence of custom prefix guidance

cormacrelf:
Ref: #487


jorgebucaran:
@cormacrelf Sorry, I asked you for a PR before I understood what were discussing in #487. Now that I understand, I realize I don't want to add this amount of detail to the README.

Users with a custom configuration should be capable to figure this out. 

The documentation suggests _User > System > Fisher_. This means Fisher packages won't be able to override system functions. But the user can still override everything.

PR #494, opened 2018-11-19
Optimize fisher startup path

neersighted:
With the addition of keybinding support for fish 2.x, fisher greatly
increases shell startup time (50-100ms in my testing). This is due to a
number of reasons, including using less-than-ideal builtins,
unconditionally running mkdir, unconditionally recreating
conf.d/fisher.fish, and checking for curl on startup (instead of when
curl is used).

This patch cuts fisher-related startup time to roughly 25ms, simplifying
the code paths as much as I am able.

The source function is removed as a compromise for speed. Fish 2.1 and
greater support `source`, and supporting fish 2.0 is not worth a hefty
speed penalty in my opinion.

curl checks have been moved to the two uses of curl -- there is some
duplication now so this may be slightly less ideal. Also, curl checks
are now done with the builtin `type` instead of `which`, which is not
guaranteed to exist.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -23,13 +23,13 @@ function fisher -a cmd -d "fish package manager"
 
     if test ! -e "$fisher_path/conf.d/fisher.fish"
         switch "$version"
-            case 2\*
+            case 2\* \*-\*
                 echo "fisher copy-user-key-bindings" > $fisher_path/conf.d/fisher.fish
         end
 
     else
         switch "$version"
-            case 2\*
+            case 2\* \*-\*
             case \*
                 command rm -f $fisher_path/conf.d/fisher.fish
         end
--- a/fisher.fish
+++ b/fisher.fish
@@ -1,7 +1,5 @@
 set -g fisher_version 3.1.1
 
-type source >/dev/null; or function source; . $argv; end
-
 switch (command uname)
     case Darwin FreeBSD
         function _fisher_now -a elapsed
@@ -14,35 +12,37 @@ switch (command uname)
 end
 
 function fisher -a cmd -d "fish package manager"
-    if not command which curl >/dev/null
-        echo "curl is required to use fisher -- install curl and try again" >&2
-        return 1
-    end
-
-    test -z "$XDG_CACHE_HOME"; and set XDG_CACHE_HOME ~/.cache
-    test -z "$XDG_CONFIG_HOME"; and set XDG_CONFIG_HOME ~/.config
+    set -q XDG_CACHE_HOME; or set XDG_CACHE_HOME ~/.cache
+    set -q XDG_CONFIG_HOME; or set XDG_CONFIG_HOME ~/.config
 
     set -g fish_config $XDG_CONFIG_HOME/fish
     set -g fisher_cache $XDG_CACHE_HOME/fisher
     set -g fisher_config $XDG_CONFIG_HOME/fisher
 
-    test -z "$fisher_path"; and set -g fisher_path $fish_config
+    set -q fisher_path; or set -g fisher_path $fish_config
 
-    command mkdir -p {$fish_config,$fisher_path}/{functions,completions,conf.d} $fisher_cache
+    for dir in {$fish_config,$fisher_path}/{functions,completions,conf.d} $fisher_cache
+        if test ! -e $dir
+            command mkdir -p $dir
+        end
+    end
 
     if test ! -e "$fisher_path/completions/fisher.fish"
         echo "fisher self-complete" > $fisher_path/completions/fisher.fish
         _fisher_self_complete
     end
 
-    if test -e "$fisher_path/conf.d/fisher.fish"
-        command rm -f $fisher_path/conf.d/fisher.fish
-    end
+    if test ! -e "$fisher_path/conf.d/fisher.fish"
+        switch "$version"
+            case 2\*
+                echo "fisher copy-user-key-bindings" > $fisher_path/conf.d/fisher.fish
+        end
 
-    switch "$version"
-        case \*-\*
-        case 2\*
-            echo "fisher copy-user-key-bindings" > $fisher_path/conf.d/fisher.fish
+    else
+        switch "$version"
+            case \*-\* 3\*
+                command rm -f $fisher_path/conf.d/fisher.fish
+        end
     end
 
     switch "$cmd"
@@ -140,7 +140,12 @@ function _fisher_self_update -a file
     set -l url "https://raw.githubusercontent.com/jorgebucaran/fisher/master/fisher.fish"
     echo "fetching $url" >&2
 
-    curl -s "$url?nocache" >$file@
+    if not type -p curl >/dev/null 2>&1
+        echo "curl is required to update fisher -- install curl and try again" >&2
+        return 1
+    end
+
+    command curl -s "$url?nocache" >$file@
 
     set -l next_version (awk 'NR == 1 { print $4; exit }' < $file@)
     switch "$next_version"
@@ -224,6 +229,11 @@ function _fisher_pkg_fetch_all
     set -l local_pkgs
     set -l actual_pkgs
     set -l expected_pkgs
+
+    if not type -p curl >/dev/null 2>&1
+        echo "curl is required to use fisher -- install curl and try again" >&2
+        return 1
+    end
 
     for id in $argv
         switch $id
--- a/fisher.fish
+++ b/fisher.fish
@@ -1,16 +1,5 @@
 set -g fisher_version 3.1.1
 
-switch (command uname)
-    case Darwin FreeBSD
-        function _fisher_now -a elapsed
-            command perl -MTime::HiRes -e 'printf("%.0f\n", (Time::HiRes::time() * 1000) - $ARGV[0])' $elapsed
-        end
-    case \*
-        function _fisher_now -a elapsed
-            command date "+%s%3N" | command awk -v ELAPSED="$elapsed" '{ sub(/%?3N$/, "000") } $0 -= ELAPSED'
-        end
-end
-
 function fisher -a cmd -d "fish package manager"
     set -q XDG_CACHE_HOME; or set XDG_CACHE_HOME ~/.cache
     set -q XDG_CONFIG_HOME; or set XDG_CONFIG_HOME ~/.config
@@ -449,5 +438,14 @@ function _fisher_wait
     while for job in $argv
             contains -- $job (_fisher_jobs); and break
         end
+    end
+end
+
+function _fisher_now -a elapsed
+    switch (command uname)
+        case Darwin FreeBSD
+            command perl -MTime::HiRes -e 'printf("%.0f\n", (Time::HiRes::time() * 1000) - $ARGV[0])' $elapsed
+        case \*
+            command date "+%s%3N" | command awk -v ELAPSED="$elapsed" '{ sub(/%?3N$/, "000") } $0 -= ELAPSED'
     end
 end
--- a/fisher.fish
+++ b/fisher.fish
@@ -40,7 +40,8 @@ function fisher -a cmd -d "fish package manager"
 
     else
         switch "$version"
-            case \*-\* 3\*
+            case 2\*
+            case \*
                 command rm -f $fisher_path/conf.d/fisher.fish
         end
     end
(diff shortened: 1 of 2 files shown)


fisher.fish
```diff
@@ -14,35 +12,37 @@ switch (command uname)
...
+                echo "fisher copy-user-key-bindings" > $fisher_path/conf.d/fisher.fish
+        end
 
-    switch "$version"
-        case \*-\*
-        case 2\*
-            echo "fisher copy-user-key-bindings" > $fisher_path/conf.d/fisher.fish
+    else
+        switch "$version"
+            case \*-\* 3\*
```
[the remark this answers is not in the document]
neersighted: Ah, fair point. fish 3 (that is, a tagged version) would not be supported by the existing logic, so I made the simplest change. Inverting the logic is trivial and more robust however, I'll go ahead and do that.

jorgebucaran:
@neersighted What if we define a `curl` function outside the `fisher` function definition in advance? Similarly to how `_fisher_now` is created. I'm not suggesting that's better, just curious.

jorgebucaran:
@neersighted I thought we could use `command -s`, but it isn't supported in 2.0. What if we don't check for `curl` at all? The user must have `curl` to have installed Fisher in the first place.

neersighted:
I considered just removing the curl checks. If you're fine with that (and curl not being found is hardly a cryptic error), then I'd prefer that path, over potentially leaking a curl wrapper function.

jorgebucaran:
@neersighted I agree. Here is another idea to keep the `curl` check, but avoid the duplication.

```fish
if command command -v curl >/dev/null
  function fisher
    echo "bad"
  end
  exit
end

function fisher
  ...
end
```

What do you think

neersighted:
We still have to pay the cost of a redirection (which is rather expensive, at around 25ms on WSL) on startup, given `command` is lacking a `-q` flag in all but the most recent tagged fish releases. From a performance standpoint I think curl should be checked at time of use, or not at all.

jorgebucaran:
@neersighted Sounds good. Then we'll proceed as you've indicated. Thank you! 🙌 

Now, this is technically a breaking change, as we're suddenly dropping fish 2.0 support, but I'll ignore that and ship this change with the 3.2 minor release that includes a large rewrite of the core functions.

FabioAntunes:
Somehow related, do we have any numbers for the amount of users that use fish 2.0

neersighted:
Sounds good to me!

faho:
>Somehow related, do we have any numbers for the amount of users that use fish 2.0

From my experience, basically nobody. Somebody's opened an issue with 2.2.0 last week (and I immediately asked them to upgrade), I can't even recall when we had the last one with a version older than that.

Going by Ubuntu versions, it's in Ubuntu 14.04 ("trusty"), which is EOL in April next year (i.e. under 6 months remaining), but is still supported by [the official PPA](https://launchpad.net/~fish-shell/+archive/ubuntu/release-2).

Similarly, we offer packages down to RHEL 5, Debian 8, openSUSE 42.2 (which is already EOL). So installing fish 2.7 on any of those is _easy_. macOS users should always have the newest version available, and FreeBSD has ports that are up-to-date.

That means there really isn't much of a need to support old fish versions. 

>Also, curl checks
are now done with the builtin type instead of which, which is not
guaranteed to exist.

For fish prior to 2.7.0, `which` was a dependency. `type` used it internally. If it doesn't exist for those, the install is broken. Since 2.7.0, it's acceptable for `which` to not exist, and those versions all have `command -sq`. Anyway, you usually want to allow functions as well, so using `type` is better anyway.

Though just doing it is also alright, as long as the error is still okay. If it just says "unknown command: curl" once, that's probably acceptable. But if you called it in a loop and you get tens of errors? Probably not.

neersighted:
I see. Thanks for the clarification, @faho.

It would also be preferable to avoid shelling out to uname during fisher startup, though I didn't see a super-elegant solution. Maybe we should just cache the result of the first `(command uname)` in a variable and move the function initialization inside `function fisher`?

jorgebucaran:
Thank you, @neersighted. Using `uname` when the file is autoloaded means `uname` is used only once, so there's nothing to win by caching it. Can you please revert your last commit?

jorgebucaran:
> @neersighted The reason for caching `uname` in `_fisher_now` is to avoid shelling out to it every time the function is invoked.

Yeah, I totally get what you are trying to do, but my understanding is that `uname` is run once (when `fisher.fish` is sourced by the shell) and not every time you invoke the `fisher` function as you are implying.

jorgebucaran:
> @neersighted The reason for caching `uname` in `_fisher_now` is to avoid shelling out to it every time the function is invoked.

Yeah, I get what you are trying to do, but my understanding is that `uname` is run once (when `fisher.fish` is sourced by the shell) and not every time you invoke the `fisher` function as you are implying.

neersighted:
Yep! The problem is that `fisher` is sourced by the shell (as part of conf.d loading) every time it starts, calling `uname` every time it starts. The caching inside of `_fisher_now` is incidental, really, and can easily be removed -- moving the `uname` call into the function allows sourcing fisher without using any non-builtins.

jorgebucaran:
@neersighted I think we have a fundamental misunderstanding here. Let's try an experiment. 

Create a `hello.fish` file in your `~/.config/fish/functions` directory as follows.


```fish
function __hello -e hello_event
  echo ONCE
end

emit hello_event

function hello
  echo HELLO
end
```

> I am using events to prevent stdout (or stderr IIRC) being swallowed by `fish` during autoload.

Now, launch a new shell and start typing `hello`. You'll see "ONCE" printed on the terminal. This happens only the first time the `hello` function is autoloaded [as you yourself explained](https://github.com/jorgebucaran/fisher/pull/494#issuecomment-440102456).

Now, try running `hello` again a few times. You should be able to confirm that "ONCE" is not printed anymore. Similarly, the code outside the `fisher` function definition (the `switch` statement using `uname` should run only once).

One more experiment. Create a snippet in `~/.config/fish/conf.d/hello.fish` that runs `hello`.

```fish
hello
```

Launch a new session. You should see "ONCE" and "HELLO" printed on the terminal. Run `hello` again. You should be able to confirm again that "ONCE" is not printed anymore.

neersighted:
I agree there's a misunderstanding -- we both agree on what's happening, but not on why that matters, I think. I'd like to avoid running `uname` during the initial sourcing in the name of avoiding non-builtin commands during shell startup, as it does contribute measurably to startup time. I'd be fine with reverting the commit if you think I'm being too picky here; however, as someone who spawns many instances of fish, I'd prefer that fisher contribute to time-to-prompt as little as possible.

jorgebucaran:
@neersighted Fair enough. Let's roll with that. Just one more thing, use `uname` every time `_fisher_now` is invoked (don't cache it).

> ...as it does contribute measurably to startup time

Please share your numbers. I'm curious to see what you see.

neersighted:
This test is hardly scientific, and I think this is a worst-case scenario because of WSL (the time to read and execute the `uname` binary from the emulated filesystem is what causes the slowness), but it appears to be 10-15ms depending on how busy my machine is. Splitting hairs maybe, but if it just takes rearranging a block of code to only use builtins, then I think it is worth it.

```
Benchmark #1: cat fisher.fish > ~/.config/fish/functions/fisher.fish; fish -i -l -c exit
  Time (mean ± σ):      55.8 ms ±   3.9 ms    [User: 13.8 ms, System: 37.1 ms]
  Range (min … max):    50.3 ms …  67.3 ms

Benchmark #2: cat fisher-old.fish > ~/.config/fish/functions/fisher.fish; fish -i -l -c exit
  Time (mean ± σ):      72.3 ms ±   3.2 ms    [User: 15.1 ms, System: 40.0 ms]
  Range (min … max):    68.3 ms …  85.3 ms

Summary
  'cat fisher.fish > ~/.config/fish/functions/fisher.fish; fish -i -l -c exit' ran
    1.29 ± 0.11 times faster than 'cat fisher-old.fish > ~/.config/fish/functions/fisher.fish; fish -i -l -c exit'
```

fisher.fish
```diff
@@ -1,48 +1,38 @@
...
+                echo "fisher copy-user-key-bindings" > $fisher_path/conf.d/fisher.fish
+        end
 
-    switch "$version"
-        case \*-\*
-        case 2\*
-            echo "fisher copy-user-key-bindings" > $fisher_path/conf.d/fisher.fish
+    else
+        switch "$version"
+            case 2\*
```
jorgebucaran: @neersighted I almost missed this. You can't remove `\*-\*`.

It means we're not handling key bindings if the user is running an unofficial fish release.

jorgebucaran:
@neersighted Fair enough, thanks for calculating that.

jorgebucaran:
@neersighted Can we just run `mkdir` and `rm` as we were before?

I think your original intention here was to eliminate Fisher's impact on shell startup time and you accomplished it. Maybe you can split this PR into two, one dealing with the changes outside the `fisher` function which I have no doubt about merging and another with the changes inside the `fisher` function, which I am not so sure about.

fisher.fish
```diff
@@ -1,48 +1,38 @@
...
-    test -z "$XDG_CONFIG_HOME"; and set XDG_CONFIG_HOME ~/.config
+    set -q XDG_CACHE_HOME; or set XDG_CACHE_HOME ~/.cache
+    set -q XDG_CONFIG_HOME; or set XDG_CONFIG_HOME ~/.config
 
     set -g fish_config $XDG_CONFIG_HOME/fish
     set -g fisher_cache $XDG_CACHE_HOME/fisher
     set -g fisher_config $XDG_CONFIG_HOME/fisher
 
-    test -z "$fisher_path"; and set -g fisher_path $fish_config
+    set -q fisher_path; or set -g fisher_path $fish_config
```
jorgebucaran: @neersighted This is fine. Please leave this. 🙏

fisher.fish
```diff
@@ -140,7 +130,7 @@ function _fisher_self_update -a file
     set -l url "https://raw.githubusercontent.com/jorgebucaran/fisher/master/fisher.fish"
     echo "fetching $url" >&2
 
-    curl -s "$url?nocache" >$file@
+    command curl -s "$url?nocache" >$file@
```
jorgebucaran: @neersighted This is fine as well. Please leave it. 🙏

fisher.fish
```diff
@@ -450,3 +440,12 @@ function _fisher_wait
         end
     end
 end
+
+function _fisher_now -a elapsed
```
jorgebucaran: @neersighted Everything below this line is perfect. 💯

fisher.fish
```diff
@@ -1,48 +1,38 @@
...
 
     set -g fish_config $XDG_CONFIG_HOME/fish
     set -g fisher_cache $XDG_CACHE_HOME/fisher
     set -g fisher_config $XDG_CONFIG_HOME/fisher
 
-    test -z "$fisher_path"; and set -g fisher_path $fish_config
+    set -q fisher_path; or set -g fisher_path $fish_config
 
-    command mkdir -p {$fish_config,$fisher_path}/{functions,completions,conf.d} $fisher_cache
+    for dir in {$fish_config,$fisher_path}/{functions,completions,conf.d} $fisher_cache
```
jorgebucaran: @neersighted The new logic is convoluted (lines 13 to 35) and I am not comfortable with it. If you can remove this from this PR and create another PR with only this, it would make my work easier.

jorgebucaran: @neersighted The new logic is convoluted (lines 13~35) and I am not too fond of it.

If you can remove that from this PR and create another PR with only that, it would make my work easier.

neersighted:
I'm not sure separating the changes makes too much sense from a performance standpoint (the entire body of the `fisher` function in run on startup, because of `fisher copy-user-key-bindings` being in a `conf.d` file), but I can certainly split it for a cleaner diff, separating out the trivial rearrangements from the logic changes.

jorgebucaran:
Keep in mind that this will be a non-issue for fish 3.0 users. If you can't simplify the code more, then leave it and I'll figure something out after merging your changes locally.

Good job @neersighted! 🙌

neersighted:
Ah, sorry I wasn't able to get around to slicing up this PR; I suppose you walked through the code and decided it was merge-ready?

Just one thing I just realized -- wouldn't `fisher copy-user-key-bindings` now run on unofficial fish 3.x builds -- is only disabling that code on official versions of fish 3 the intention?

jorgebucaran:
@neersighted No problem. I found it easier to merge your changes in my working branch for 3.2.0.

> ...now run on unofficial fish 3.x builds

`fisher copy-user-key-bindings` should only run for official fish 2.x builds. If that's not the case, then I missed it, but don't worry, I'll fix it.

Thank you for working on this! 🎉

PR #508, opened 2019-01-11
add gawk dependency

edouard-lopez:
Running `fisher` in a docker with busybox I had trouble due to the version available (`awk` not `gawk`). 

The situation was solved by running:

    apk add gawk

Not sure about the minimal version


jorgebucaran:
@edouard-lopez Turns out this was a bug in https://github.com/jorgebucaran/fisher/commit/fff873955a8eadd73fb074bb0b9df8bb726155e0. Thanks.

PR #516, opened 2019-01-18
Override fishfile content instead of replacing the file

stefanmaric:
This allows fishfile to be a symlink, which I used to keep track of my
dotfiles in a repo, and broke after upgrading to V3.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -240,8 +240,7 @@ function _fisher_commit -a cmd
         end
     end
 
-    _fisher_fmt <$fishfile | _fisher_write $cmd $actual_pkgs >$fishfile.
-    command mv -f $fishfile. $fishfile
+    printf "%s\n" (_fisher_fmt <$fishfile | _fisher_write $cmd $actual_pkgs) >$fishfile
 
     _fisher_self_complete
 


jorgebucaran:
@stefanmaric What about a one-liner like this—would that work too?

```fish
printf "%s\n" (_fisher_fmt <$fishfile | _fisher_write $cmd $actual_pkgs) >$fishfile
```

stefanmaric:
It does work. I guess `printf` doesn't stream and instead holds all the data in memory. :smile:

jorgebucaran:
Let's go with this! I'll merge this now, but bump the version later! Thanks.  💯 

> I thought the intermediary file was to prevent errors while streaming the file and writing to it at the same time.

I found it more readable that way and that's the only reason. We were not checking for errors either.

After your changes, reading the file will occur in a subshell and `printf` will not get to it until we've finished parsing the output. When `printf` runs there will be just an array of entries to write to the fishfile.

stefanmaric:
> After your changes, reading the file will occur in a subshell and `printf` will not get to it until we've finished parsing the output. When `printf` runs there will be just an array of entries to write to the fishfile.

You are totally right. That was actually my initial thought after your code suggestion, but I wrongly ditched that idea because of the fact `fish` doesn't have subshells (`echo (cd /; and ls)`, for example, will move you out of your current dir because the shell expression is executed in your current shell, whereas `echo "$(cd / && ls)"` in `bash` will not).

It isn't a subshell, but the command substitution does run through completion before `printf` being called.

Thanks for the insight!

jorgebucaran:
> It isn't a subshell, but the command substitution does run through completion before `printf` being called.

I stand corrected. Edited my answer as I should've said command substitution and _not_ subshell. 

I was thinking about execution order and meant that `printf` would receive the result of the command substitution as an argument!

> Thanks for the insight!

Thank you! 🙌

jorgebucaran:
> It isn't a subshell, but the command substitution does run through completion before `printf` being called.

I stand corrected. Edited my answer as I should've said command substitution and _not_ subshell. 

I was thinking about execution order and meant that `printf` would receive the result of the command substitution as an argument.

> Thanks for the insight!

Thank you! 🙌

PR #535, opened 2019-04-17
docs: match version and help commands to actual

will-stone:
The help and version commands use the "--" prefix. This updates the readme to reflect that.


jorgebucaran:
Thanks! 🙌

PR #537, opened 2019-05-26
Detect when $fisher_path = $fisher_config and exit with error message

9999years:
The fishfile is stored at `$fisher_path/fishfile`, and `_fisher_commit` executes `rm -rf $fisher_config`, so if `$fisher_path = $fish_config`, running `fisher` will delete the fishfile before looking for packages, leading to confusing errors.

Adds a check to `fisher`, exiting with a message like

    $fisher_path (/home/arvensis/.config/fisher) cannot be /home/arvensis/.config/fisher
    Consider instead:
        set -g fisher_path "$HOME/.config/fisher_local"

To prevent user confusion in the future.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -16,6 +16,17 @@ function fisher -a cmd -d "fish package manager"
         end
     end
 
+    # The fishfile is stored at $fisher_path/fishfile, and _fisher_commit
+    # executes `rm -rf $fisher_config`, so if $fisher_path = $fish_config,
+    # running `fisher` will delete the fishfile before looking for packages,
+    # leading to confusing errors.
+    if test (readlink "$fisher_path") = (readlink "$fisher_config")
+        echo "Error: \$fisher_path ($fisher_path) cannot be $fisher_config"
+        echo "Consider instead:"
+        echo "    set -g fisher_path \"\$HOME/.config/fisher_local\""
+        return 1
+    end
+
     if test ! -e $fisher_path/completions/fisher.fish
         echo "fisher complete" >$fisher_path/completions/fisher.fish
         _fisher_complete


jorgebucaran:
Thank you, @9999years, I'm aware of this PR, I just haven't had time to review it properly. I'll have a look at this soon and give you my feedback. 👋

fisher.fish
```diff
@@ -16,6 +16,17 @@ function fisher -a cmd -d "fish package manager"
         end
     end
 
+    # The fishfile is stored at $fisher_path/fishfile, and _fisher_commit
+    # executes `rm -rf $fisher_config`, so if $fisher_path = $fish_config,
+    # running `fisher` will delete the fishfile before looking for packages,
+    # leading to confusing errors.
+    if test (readlink "$fisher_path") = (readlink "$fisher_config")
```
jorgebucaran: @9999years I don't believe using `readlink` is cross-Unix compatible?

jorgebucaran:
@9999years I see what's going on here. I suppose you run into this, hence the desire to add a check, but I suspect `$fisher_path` being equal to `$fish_config` is an oddity. Can't back up my claim, though, so who knows?

For starters, we 'd need to switch from `readlink` to something compatible across Unices. The  `realpath` builtin is the easy way out, but that means dropping support for a bunch of older fish. 

On the other hand, a comment in the docs might be the next best thing to do here. Thoughts?

reitzig:
> I suspect `$fisher_path` being equal to `$fish_config` is an oddity. Can't back up my claim, though, so who knows?

Don't you [always set it](https://github.com/jorgebucaran/fisher/blob/master/fisher.fish#L11), though?

> `set -q fisher_path; or set -g fisher_path $fish_config`

Not sure how `$fisher_path` could have been set already at that point. But then, I also don't understand why that line makes sense, so maybe I'm missing something fundamental here.

jorgebucaran:
Hi @9999years, sorry for the late reply. Would you mind checking if this is still issue in fisher@[`3.2.12`](https://github.com/jorgebucaran/fisher/releases/tag/3.2.12)?

jorgebucaran:
Not going to merge here, but I'll revisit this issue when I start working on #582.

Issue #539, opened 2019-06-28
custom fish_greeting overwrite after fisher update all package


kenchou:
* install plugin
  * laughedelic/fish_logo
  * oh-my-fish/theme-bobthefish

* add fish_logo to ~/.config/fish/functions/fish_greeting.fish
```
function fish_greeting
    type -q fish_logo; and fish_logo
end
```
* and start new session will see the fish logo
* after execute `fisher`, the `~/.config/fish/functions/fish_greeting.fish` has been changed (provide by theme-bobthefish), and my greeting was lost.
```
function fish_greeting -d "What's up, fish?"
    set_color $fish_color_autosuggestion
    uname -nmsr

    # TODO: `command -q -s` only works on fish 2.5+, so hold off on that for now
    command -s uptime >/dev/null
    and uptime

    set_color normal
end
```

jorgebucaran:
@kenchou That's what I expect to happen when I install `bobthefish` since `fish_greeting.fish` is part of that package.

If you want to install `bobthefish` but preserve your previous `fish_greeting.fish`, you could customize Fisher's installation prefix to a different location using the `$fisher_path` env variable as described in the docs.

Odaeus:
I just came here to open a very similar issue. I installed Spacefish via Fisher and it overwrote my `fish_prompt` files 😞. It was totally unexpected as no package manager I'm familiar with would behave in that way. For example, `apt` will ask whether files should be overwritten or Vim's Pathogen has each package in its own directory.

@jorgebucaran Can I please ask you to really add a specific warning to the README stating that Fisher will overwrite files if they are contained in a package? I know there is wording under "Changing the installation prefix" similar to this but it reads like the focus is simply on personal directory organisation rather than such an important caveat.

jorgebucaran:
@Odaeus Why did you think was going to happen to your `~/.config/fish/fish_prompt.fish` after you installed Spacefish?

Odaeus:
@jorgebucaran I didn't think about the internal workings of Fish or Fisher at all. Obviously I expected my prompt config to be superseded in some way, but my surprise was that it destroyed the original data. If these were files that users aren't supposed to touch, then sure, but my understanding is that these are user-controlled files. I suppose I expected perhaps a prompt asking "overwrite this file?" before it proceeded, or that Fisher packages were kept separate and used path wrangling to override user-defined functions.

Odaeus:
[...]

PR #541, opened 2019-07-20
Add note about how packages are installed

Odaeus:
Following on from discussion in https://github.com/jorgebucaran/fisher/issues/539.

This adds a section to the README that aims to prevent users from possibly losing data if they are playing around with Fisher and are unaware of how it works.

Hopefully this is clear, to the point, and prominent enough even though I avoided putting **WARNING** 😄.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,11 @@ if not functions -q fisher
 end
 ```
 
-### Changing the installation prefix
+### Setting the installation path
 
-Use the `$fisher_path` environment variable to change the location where functions, completions, and configuration snippets will be copied to when a package is installed. The default location will be your fish configuration directory.
+By default, fisher expands packages into the fish configuration directory, which may cause existing customized files (such as `fish_prompt.fish`) to be overwritten. Use the `$fisher_path` environment variable to change the location where functions, completions, and configuration snippets will be copied to when a package is installed.
 
-> **Note**: Do I need this? If you want to keep your own functions, completions, and configuration snippets separate from packages installed with Fisher, customize the installation prefix. If you prefer to keep everything in the same place, you can skip this.
+> **Note**: Do I need this? If you want to keep your own functions, completions, and configuration snippets separate from packages installed with Fisher, customize the installation path. If you prefer to keep everything in the same place, you can skip this.
 
 ```fish
 set -g fisher_path /path/to/another/location


jorgebucaran:
Thanks, @Odaeus. I think we can simplify the text and merge it into the "Changing the installation prefix" section. What do you think?

Odaeus:
I considered whether it belonged in that section but the problem for me is that "Changing the installation prefix" sounds like something unusual that advanced users would want to do. If I was skimming the document, I would see the title and skip that section thinking "nope, why would I do that" (erroneously based on assumptions on how Fisher works).

If you are happy for me to change the title "Changing the installation prefix" to something like "Package installation location" then I think people are more likely to pay attention to it and I can merge the sections?

jorgebucaran:
I think https://github.com/jorgebucaran/fisher/pull/554 was a dupe here and I didn't realize. @Odaeus Do you think that'll be enough or would you like us to elaborate more?

Odaeus:
Thanks for following up @jorgebucaran. I've looked at #554 but I'm not sure I would consider it a dupe. If I understand correctly, that update is warning against using the `fisher` dir for storing custom config. My proposal is about informing users that by default, Fisher packages can overwrite existing Fish config files without warning.

If you think my current text is too strong as a standalone section, I would suggest:
* Changing the heading of "Changing the installation prefix" to "Package files location"
* Starting the paragraph with my first sentence: "By default, Fisher expands packages directly into the Fish config directory, which may cause existing files to be overwritten."

jorgebucaran:
> @Odaeus Fisher packages can overwrite existing Fish config files without warning.

Is this true?

Odaeus:
@jorgebucaran ah, I don't know what the correct term is? I refer to `.config/fish` as "fish config directory". But actually most of those files are not under user control. If it is more correct to say "By default, Fisher expands packages directly into the Fish *functions* directory, which may cause existing files to be overwritten." then that's also fine. I don't know if there are additional directories (e.g. `completions`) that packages can write to.

jorgebucaran:
@Odaeus Okay, let's go with the shorter version. If you can find a place within the existing sections to add that, I'll merge this in a heartbeat! 💯

README.md
```diff
@@ -42,9 +42,9 @@ if not functions -q fisher
 end
 ```
 
-### Changing the installation prefix
+### Package files location
 
-Use the `$fisher_path` environment variable to change the location where functions, completions, and configuration snippets will be copied to when a package is installed. The default location will be your fish configuration directory.
+By default, fisher expands packages into the fish configuration directory, which may cause existing customized files (such as `fish_prompt.fish`) to be overwritten. Use the `$fisher_path` environment variable to change the location where functions, completions, and configuration snippets will be copied to when a package is installed.
```
jorgebucaran, line 47: @Odaeus Perfect. 👍

README.md
```diff
@@ -42,9 +42,9 @@ if not functions -q fisher
 end
 ```
 
-### Changing the installation prefix
+### Package files location
```
jorgebucaran, line 45: @Odaeus I prefer the original title! 🙏

Odaeus: I appreciate that and I addressed it in my [second comment at the top of this page](https://github.com/jorgebucaran/fisher/pull/541#issuecomment-513539568). It negates the purpose of my effort to document this problem if a reader will most likely skip over it because the heading sounds like an esoteric action that they would have no interest in performing.

Please close this PR if you do not wish to change the heading as I don't see another way 
to make it prominent enough to help new users avoid data loss.

jorgebucaran: I'll add your changes anyway, but I am open meet you in the middle with a different title. I agree that "Changing the installation prefix" sounds "advanced". What about "Changing the installation path" or "Customizing the installation path". They don't fully address your original intent, but sound a bit less intimidating.

Odaeus: "Prefix" does indeed make it sound advanced and you're right that "path" is a more natural choice, especially as it matches the variable name. Though I still believe many people would skip over the section with a verb like "change" or "customize" in the belief that they would have no reason to change the default. Whereas a heading with a noun phrase appears more informational.
As one last attempt, I suggest the verbs "setting" or "managing" as both sound more equivocal on whether it's something that should be changed or not.

Issue #545, opened 2019-07-30
Make path for fishfile configurable to be separate from $fisher_path


mattmc3:
I keep my fish configuration in my dotfiles repo, but don't really need to keep all stuff installed via fisher there too, just the `fishfile`. Problem is, if you change `$fisher_path`, it changes where `fishfile` gets placed to store it in $fisher_path. I would argue that your fishfile is a separate concept than what gets installed, so fishfile should be configurable to live with the rest of your config.

So, for example, if I set `$fisher_path` to `$XDG_DATA_HOME/fish/fisher`, `fishfile` gets placed in `$fisher_path` instead of `~/.config/fish/fishfile`.

I could symlink fishfile so it gets committed, but that seems like a bad solution. I saw [this issue](https://github.com/jorgebucaran/fisher/issues/524) and wondered under what conditions you would accept a PR that made both these changes? I'm willing to put in the work to do the following:

1. Change the name from fishfile to fish_bundle
2. Add an optional variable to set $fish_bundle_path, and change the old default location from $fisher_path/fish_bundle to $fish_config/fish_bundle
3. Migrate existing `fishfile`s to the new `fish_bundle`. You'd probably have to leave this in for a few versions while people update.

Let me know your thoughts.

jorgebucaran:
@mattmc3 Have you read https://github.com/jorgebucaran/fisher#changing-the-installation-prefix?

jorgebucaran:
@mattmc3 Have you read [Changing the installation prefix](https://github.com/jorgebucaran/fisher#changing-the-installation-prefix)? :)

mattmc3:
Hi Jorge, thanks for replying. I did read that and have changed my installation prefix by setting fisher_path. I think maybe you misread my request, or maybe I’m missing a part you’re referring to?

Setting the fisher_path variable also unfortunately has the side effect of changing the location of fishfile, which I believe is undesirable. I would like to be able to configure the location of fishfile separately or in addition to the location of fisher_path with the introduction of a new fisher_bundle_path variable. Did you see my PR?

jorgebucaran:
@mattmc3 Could you have a look at https://github.com/jorgebucaran/fisher/issues/479 & https://github.com/jorgebucaran/fisher/pull/480? They're related. I can't recall now, but I believe that's how I initially implemented Fisher.

> Let me know your thoughts.

[...]

Issue #524, opened 2019-01-31
Rename fishfile to fish_plugins


jorgebucaran:
Merely for aesthetic consistency. There is already a `fish_history` and `fish_variables` file in your `~/.config/fish` directory. Renaming `fishfile` to ~~`fish_bundle`~~ (or ~~`fish_file`~~) **`fish_plugins`** would be nice. 

```console
$ la ~/.config/fish
jb  staff   320B Jan 31 00:51 .
jb  staff   448B Jan 31 00:48 ..
jb  staff   128B Jan 31 00:48 completions
jb  staff   128B Jan 31 00:48 conf.d
jb  staff   190K Jan 28 18:31 fish_history
jb  staff   1.3K Jan 31 00:51 fish_variables
jb  staff    22B Jan 31 00:48 fish_bundle
jb  staff   256B Jan 31 00:48 functions
```

ammgws:
What about `fisher_bundle`? At least to make it obvious where the file comes from.

patrickf3139:
Besides aesthetic consistency, I also like either `fish_bundle` or `fisher_bundle` because fishfile could refer to so many different things. I mean, the name only implies a file having to do with fish.

mattmc3:
I'm willing to submit a PR for this if we can agree on the naming @jorgebucaran. I'm partial to fish_bundle rather than fisher_bundle since it's really just a list of fish plugins and not anything with fisher specific data or formatting, so it could even be used by other fish plugin management tools, but honestly, whatever you want is fine.

My questions are: Is this still something we want to do, and are there any other requirements other than simply:

1. Change the name of fishfile to fish_bundle
2. Handle migration of fishfile to fish_bundle if the former exists, but not the latter

jorgebucaran:
@mattmc3 I like `fish_bundle` too. It just looks cool. I get that `fisher_bundle` would be a safer approach, but then we might as well leave things the way they are.

I'm happy to go with this if we can agree on `fish_bundle`.

/cc @franciscolourenco @edouard-lopez @FabioAntunes @Scrumplex @neersighted 

> ...are there any other requirements other than simply:

[...]

PR #546, opened 2019-07-31
fishfile to fish_bundle and config variable

mattmc3:
1. Convert from using fishfile to fish_bundle: https://github.com/jorgebucaran/fisher/issues/524
2. Add optional variable ($fisher_bundle_path) for custom storage location of fish_bundle: https://github.com/jorgebucaran/fisher/issues/545
3. Add .editorconfig file so that contributors can have their environment set up properly for this project's code style
4. Bump version to 3.3.0

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+[Makefile]
+indent_style = tab
+
+[*.fish]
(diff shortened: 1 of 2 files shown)


mattmc3:
I should update the README as well detailing the new fisher_bundle_path variable in the “Changing the installation prefix” section. It might also be worth a quick discussion about whether this variable should contain the directory (as written), or the whole path to the fishfile (fish_bundle now). We could allow both and detect too.

jorgebucaran:
> @mattmc3 I should update the README as well detailing the new fisher_bundle_path variable in the “Changing the installation prefix” section.

Yes, but before we do that, let's discuss in #545. 🙏

mattmc3:
Closing in favor of a different implementation.

PR #571, opened 2020-06-24
Migrate Data to Proper XDG Directory

### LINKED ISSUE #568: Migrate fisher dir from XDG_CONFIG_HOME to XDG_DATA_HOME
To show case what I mean, I am going to paste the information another tool called direnv does in a recent release:

> Migrating the allow data to the new location
>
> The allowed .envrc permissions used to be stored in the XDG_CONFIG_HOME. It's
better to keep that folder for user-editable configuration so the data is
being moved to XDG_DATA_HOME.
>
> moving /home/userA/.config/direnv/allow to /home/userA/.local/share/direnv/allow
creating a symlink back from /home/userA/.local/share/direnv/allow to /home/userA/.config/direnv/allow for back-compat.
>
> All done, have a nice day!

As such, it would make sense for fisher to also migrate from `XDG_CONFIG_HOME` to `XDG_DATA_HOME`. The config is used for user-editable config whereas the data dir should be used for data written by the program such as fisher.

One of many reasons to do this is that a lot of users utilize dotfile support to backup configuration data, and that is often done for the `XDG_CONFIG_HOME` dir.

See also https://github.com/jorgebucaran/fish-nvm/issues/108

james-r-smith:
Going by the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), it makes more sense to place fisher data in `XDG_DATA_HOME` rather than `XDG_CONFIG_HOME` based on fisher's current usage of the directory. This PR covers the following:

* Swap all instances of `fisher_config` with `fisher_data`, which will point to `XDG_DATA_HOME`
* Run a check (until new major version is introduced), which will perform the directory migration

I chose to keep the README as is, as most users should still avoid placing their data in `XDG_CONFIG_HOME/fisher` in case fisher will need it in the future for other reasons.

This will close #568.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -47,6 +47,7 @@ function fisher -a cmd -d "fish package manager"
     # 2020-06-23: temp code, migrates fisher data from XDG_CONFIG_HOME to XDG_DATA_HOME
     set -l fisher_config $XDG_CONFIG_HOME/fisher
     if test -d $fisher_config
+        echo "migrating local data from $fisher_config to $fisher_data"
         command rm -rf $fisher_data
         command mv -f $fisher_config $fisher_data
     end
--- a/fisher.fish
+++ b/fisher.fish
@@ -3,10 +3,11 @@ set -g fisher_version 3.2.11
 function fisher -a cmd -d "fish package manager"
     set -q XDG_CACHE_HOME; or set XDG_CACHE_HOME ~/.cache
     set -q XDG_CONFIG_HOME; or set XDG_CONFIG_HOME ~/.config
+    set -q XDG_DATA_HOME; or set XDG_DATA_HOME ~/.local/share
 
     set -g fish_config $XDG_CONFIG_HOME/fish
     set -g fisher_cache $XDG_CACHE_HOME/fisher
-    set -g fisher_config $XDG_CONFIG_HOME/fisher
+    set -g fisher_data $XDG_DATA_HOME/fisher
 
     set -q fisher_path; or set -g fisher_path $fish_config
     set -g fishfile $fish_config/fishfile
@@ -41,6 +42,13 @@ function fisher -a cmd -d "fish package manager"
     # 2019-10-22: temp code, migrates fishfile from old path back to $fish_config
     if test -e "$fisher_path/fishfile"; and test ! -e "$fishfile"
         command mv -f "$fisher_path/fishfile" "$fishfile"
+    end
+
+    # 2020-06-23: temp code, migrates fisher data from XDG_CONFIG_HOME to XDG_DATA_HOME
+    set -l fisher_config $XDG_CONFIG_HOME/fisher
+    if test -d $fisher_config
+        command rm -rf $fisher_data
+        command mv -f $fisher_config $fisher_data
     end
 
     switch "$cmd"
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Thank you, @james-r-smith. @thernstig Is this what you were asking for?

fisher.fish
```diff
@@ -43,6 +44,13 @@ function fisher -a cmd -d "fish package manager"
         command mv -f "$fisher_path/fishfile" "$fishfile"
     end
 
+    # 2020-06-23: temp code, migrates fisher data from XDG_CONFIG_HOME to XDG_DATA_HOME
+    set -l fisher_config $XDG_CONFIG_HOME/fisher
+    if test -d $fisher_config
+        command rm -rf $fisher_data
```
thernstig, line 50: How about a courtesy notice here to the user to let them be aware this happens?

james-r-smith: Nice! I'll get that in right away. 😊 👍

james-r-smith: @thernstig & @jorgebucaran, I added the notice. I was thinking about adding a note that it will / should happen only once, but after running another command, the user should be aware that it was only a one-time call. Let me know what you think. 🙂

thernstig: @jorgebucaran decides if he likes it or not :)

james-r-smith: What do you think, @jorgebucaran? Is the message good now, or do you want a note saying the migration will only happen once?

jorgebucaran: Thank you @thernstig & @james-r-smith. The message is fine, thank you.

james-r-smith:
@jorgebucaran, any reason the PR hasn't been merged yet? Let me know and I'll address it as soon as I get a chance to. 😊 👍

jorgebucaran:
My bad, @james-r-smith. Thank you for your work on this. 👋

jorgebucaran:
@thernstig @james-r-smith I should've asked before, but what about fish's own usage of the XDG_ directory spec. They use XDG_CONFIG_HOME, not XDG_DATA_HOME.

thernstig:
fish is using both `XDG_DATA_HOME` and `XDG_CONFIG_HOME`.

They place user config in `XDG_CONFIG_HOME`, and generated (a.k.a. "data files) in `XDG_DATA_HOME`.

The data moved in this PR properly places fish data in `XDG_DATA_HOME`.
If fisher has user-config, then it should surely be placed in `XDG_CONFIG_HOME`.

Is there any configuration users edit manually for fisher?

jorgebucaran:
> Is there any configuration users edit manually for fisher?

The closest I can think of is the `fishfile` which is like `package.json`, but that's not considered configuration, right? Other than that, there is an environment variable users can customize (`$fisher_path`). There's nothing else.

thernstig:
Since `fishfile` is in the fish config directory, which already uses `XDG_CONFIG_HOME` (i.e. it defaults to `~/.config/fish/fishfile`), that was not affected by this commit. I would kind of consider it configuration if you expect/believe users should edit it manually. I think you would be the better judge if you consider that "user config"  or not?

One could possibly argue it should be in `$XDG_CONFIG_HOME/fisher/fishfile` - but that is another issue that you might not be as comfortable changing. I'd _guess_ that the fish authors would prefer that though as it has nothing to do with the fish config, but at the same time the contents of it will affect anything in the fish dir. So maybe that guess is incorrect 😆

jorgebucaran:
> I'd guess that the fish authors would prefer that though as it has nothing to do with the fish config, but at the same time the contents of it will affect anything in the fish config dir.

Exactly. The `fishfile` is at `$fisher_path/fishfile`, which is only `$XDG_CONFIG_HOME/fish` by default. 

I think we're good then! 🎉

jorgebucaran:
Thank you @james-r-smith! [**3.2.1**](https://github.com/jorgebucaran/fisher/releases/tag/3.2.12) is now officially out! 🦐

Issue #527, opened 2019-02-23
Update specific packages without re-installing everything


jorgebucaran:
Let's add a new command to update packages: `fisher update <package...>`.

Why? Sometimes you want to update one or more packages without re-installing everything. One of such cases is when developing a local package. After doing it a few times, even waiting a few milliseconds can be annoying. 

Using `update` will download only the specified packages and their dependency trees, reducing the total number of network requests. Updating a package should not trigger a <code><i>pkg</i>_uninstall</code> event (#526). 

Still on the table is whether we should emit a new event like <code><i>pkg</i>_update</code>. Should we also emit a <code><i>pkg</i>_install</code> event for newly added packages? 

I'm not too keen on adding new events, but maintaining consistency is one argument in their favor. ~~If I decide to add them, they'll probably come in a future minor release after 3.3.0.~~ Coming in 4.0.

tsdh:
Will that also allow for a more safe update mode?  The single complaint I have with just running `fisher` is that it leaves you with a broken setup in case you run it on a wonky network where deleting everything works but fetching and reinstalling won't.  It would be much better if it would first download to a temporary location and then remove and install.

Also a question related to this issue: it would be interesting to know if an update was a real update or the same version has simply been reinstalled.

jorgebucaran:
> It would be much better if it would first download to a temporary location and then remove and install.

Hmm, interesting! Thanks for the idea.

> it would be interesting to know if an update was a real update or the same version has simply been reinstalled.

I agree that it would, but that's not part of my plan this time. Thanks for the suggestion, though!

jorgebucaran:
Implemented in [4.x](https://github.com/jorgebucaran/fisher/releases/tag/4.0.0). :tada:

PR #573, opened 2020-06-25
Install event support

joseluisq:
Following the same aproach as the `uninstall` event, this PR adds support for a install event as well.
This event can be used following `file_name_install` event nomenclature.

Example:

```fish
function foobar_install -e foobar_install
  echo "Installing foobar..."
end
```

In this case, `foobar_install` function will get called when the `foobar` package gets installed via `fisher add foobar`.

I have also refactored the documentation under the "Configuration snippets" section.

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -218,9 +218,21 @@ function _fisher_commit -a cmd
     if test "$cmd" = "rm"
         set out_pkgs $next_pkgs
     else
+        set -l current_pkg_name (command basename $argv)
+        set -l current_pkg_name_path $fisher_path/conf.d/$current_pkg_name.fish
+
         for pkg in $next_pkgs
             if contains -- (echo $pkg | command sed "s|@.*||") $actual_pkgs
                 set out_pkgs $out_pkgs $pkg
+
+                set -l pkg_name (command basename $pkg .fish)
+
+                if test "$pkg_name" = "$current_pkg_name"
+                    set -l target (command basename $current_pkg_name_path)
+                    set -l filename (command basename $target .fish)
+
+                    test "$filename.fish" = "$target"; and emit "$filename"_install
+                end
             end
         end
     end
--- a/README.md
+++ b/README.md
@@ -256,7 +256,23 @@ You can leave off the `github.com` part of the URL when adding or removing packa
 
 Configuration snippets consist of all the fish files inside your `~/.config/fish/conf.d` directory. They run on [shell startup](http://fishshell.com/docs/current/index.html#initialization) and generally used to set environment variables, add new key bindings, etc.
 
-Unlike functions or completions which can be erased programmatically, we can't undo a fish file that has been sourced without creating a new shell session. For this reason, packages that use configuration snippets provide custom uninstall logic through an uninstall [event handler](https://fishshell.com/docs/current/#event).
+Unlike functions or completions which can be erased programmatically, we can't undo a fish file that has been sourced without creating a new shell session. For this reason, packages that use configuration snippets provide custom install or uninstall logic through [event handlers](https://fishshell.com/docs/current/#event)
+
+#### Install event
+
+Fisher will emit an install event if you define a package's function following the `file_name_install` event nomenclature.
+
+Example:
+
+```fish
+function foobar_install -e foobar_install
+    echo "Installing foobar..."
+end
+```
+
+`foobar_install` function will get called when you install your package via `fisher add foobar`.
+
+#### Uninstall event
 
 Let's walk through an example that uses this feature to add a new key binding for the Control-G sequence. Let's say we want to use it to open the fishfile in the `vi` editor quickly. When you install the package, `fishfile_quick_edit_key_bindings.fish` will be sourced, adding the specified key binding and loading the event handler function. When you uninstall it, Fisher will emit an uninstall event.
 


jorgebucaran:
Hi @joseluisq. What's the use case here?

joseluisq:
> What's the use case here?

@jorgebucaran when someone wants to perform a specific task once (during a package installation only).

For instance, I want to download a third-party file (E.g a config file than I might update on every release) but only when users install the latest package.

Actually I found no way to do this.

fisher.fish
```diff
@@ -218,9 +218,21 @@ function _fisher_commit -a cmd
...
+        set -l current_pkg_name (command basename $argv)
+        set -l current_pkg_name_path $fisher_path/conf.d/$current_pkg_name.fish
+
         for pkg in $next_pkgs
             if contains -- (echo $pkg | command sed "s|@.*||") $actual_pkgs
                 set out_pkgs $out_pkgs $pkg
+
+                set -l pkg_name (command basename $pkg .fish)
+
+                if test "$pkg_name" = "$current_pkg_name"
```
joseluisq, line 230: @jorgebucaran basically here I'm checking for emitting just once the event during installation.

Using your same `uninstall` approach on `_fisher_rm` function.

https://github.com/jorgebucaran/fisher/blob/c2b3ab01713256324720ee24ff2aa2da779a138d/fisher.fish#L399-L401

jorgebucaran: Will this actually run only once? What happens when I run `fisher` again? See https://github.com/jorgebucaran/fisher/issues/527

joseluisq: > Will this actually run only once?

Yes, when I say once I mean on every `fisher add foobar` only. 
Additionally, for example for new sessions created opening a new terminal tab, this event never gets emitted.

My concern was that I have no found a solution to make stuff on just every `fisher add foobar`.
Unless you have another more proper way to do it.

>  What happens when I run fisher again?

I have tried many times and I didn't see issues. I only saw that the `uninstall` event gets emitted at the beginning (related to #527), but then the normal `install` event gets emitted just at the end of the installation.

I have tested this PR with my package on local which has both events defined. Then performing `fisher add $PWD` many times as needed.

>  See #527

Is that already implemented or in progress?

patrickf3139:
Hi all,
As a package author, I am very interested in this as well. My use case will be to symlink a git-related function being installed by fisher into the user's `bin`, or at least print out a message advising the user to do so, so that the functions can be executed by `git`. e.g. if you put function called `git-command` in the user's path, then `git command` will execute `git-command`. Unfortunately, there's no clean way to do that right now.

@joseluisq if I may make a suggestion, I think a cleaner way to execute only on package install is to through an `install.fish` similar to how there is a `uninstall.fish`. It's so clean and simple thanks to the very elegant way @jorgebucaran designed `fisher`. See what I mean at https://github.com/jorgebucaran/fisher/blob/master/fisher.fish#L382.

jorgebucaran:
@patrickf3139 `"$pkg_name"_install.fish` you mean, right?

patrickf3139:
I mean the file, not the function that is triggered by the uninstall event. Here is an example https://github.com/patrickf3139/open_github/blob/master/functions/uninstall.fish.

jorgebucaran:
@patrickf3139 Using an event is probably better.

@joseluisq Would it be possible to achieve what you are trying to do using a conf.d snippet? Just want to make sure we're not reinventing the wheel here.

patrickf3139:
@jorgebucaran Would you mind explaining why an event in a conf script is better than a `install.fish` file?

patrickf3139:
It just seems to be that a file that only has to be read once when the plugin is installed vs a function that has to be re-read (though not executed most of the time) on every single fish invocation might be more elegant and efficient.

joseluisq:
> @joseluisq Would it be possible to achieve what you are trying to do using a conf.d snippet? Just want to make sure we're not reinventing the wheel here.

@jorgebucaran can you provide me an example on how to do so?

> @patrickf3139 Using an event is probably better.

@jorgebucaran I also support the Fish _event emit_ functionality.
@patrickf3139 an `install.sh` file can be an option too but Fish already provides [event handlers](https://fishshell.com/docs/current/#event) so I support the idea of using an install/uninstall events.
You can easily [source](https://fishshell.com/docs/current/cmds/source.html) a `install.sh` file within a function if you want too.

patrickf3139:
I see, is the reason to use event handlers because fish already provides support for them, so it's less logic to have to include and maintain in fisher?
I still slightly prefer `install.fish` because it seems to me that `conf.d` is for configuration, whereas this is for installation, so it's poor organization of logic. However, I don't think it's a big deal and will defer the design decisions to Jorge. He has already made many brilliant, well thought out design decisions with fisher. 
Thank you both!

IlanCosman:
I would also have a use for this feature in Tide where I want to offer the user the configuration wizard on first install but not on every shell startup. One could do this using a conf.d snippet that removes itself, but that seems unnecessarily awkward.

jorgebucaran:
@joseluisq Switching to `main` and deleting `master` closed this PR. My bad José! 😅  

You may create a new PR against the repo's new default branch, but I'd probably hold that off for now. I'll look into this when I start working on #582, at which point I'll probably rewrite this part entirely. See that Install, Update, and Uninstall event support will be introduced in one form or another in Fisher 4. 🎉

patrickf3139:
Did you switch from master to main to make the language more inclusive? If so, 👏

jorgebucaran:
Definitely!

joseluisq:
> @joseluisq Switching to `main` and deleting `master` closed this PR. My bad José! 
> 
> You may create a new PR against the repo's new default branch, but I'd probably hold that off for now. I'll look into this when I start working on #582, at which point I'll probably rewrite this part entirely. See that Install, Update, and Uninstall event support will be introduced in one form or another in Fisher 4.

I'm wondering whether you just closed this because of branch renaming or some future v4, what happen with the stable v3?
Do I have to wait for a next major release in order to use this feature? If so, I don't support your aproach. The PR has been made for the current stable release.
Anyway it's your decision as maintainer. I don't question it.

However as a result of this approach, that I don't share, I'm going to stop resending any PR, leaving this to someone else if there is some interest on it.

jorgebucaran:
Sorry for the great delay, but this is now finally implemented in [4.x](https://github.com/jorgebucaran/fisher/releases/tag/4.0.0). :tada:

cc @joseluisq @patrickf3139

PR #580, opened 2020-07-27
self-update: better network error handling

### LINKED ISSUE #578: self-update: Add ability to fail unless we get a HTTP OK response from GitHub?
Funny thing, I tried to update fisher some minutes ago, while GitHub was going through a [power outage](https://twitter.com/githubstatus/status/1282575172223016960).

After I ran `fisher self-update`, my `fisher.fish` script ended up being overwritten by this:

```
$ cat ~/.config/fish/functions/fisher.fish
500: Internal Server Error
```

I've already fixed it, but it would be nice to prevent this in the future... 😄

guihkx:
By using curl's `-f` option, we improve the self-update mechanism by not overwriting the fisher script if the HTTP status code given to us by Github is higher than 400.

Additionally, we make sure that the exit code of the curl process must be always zero for the self-update to succeed.

The `-S` option, when used with `-s`, allows us to get an error message from curl when it fails.

A few examples:

image

Fixes #578

### DIFF
--- a/fisher.fish
+++ b/fisher.fish
@@ -155,26 +155,16 @@ end
 function _fisher_self_update -a file
     set -l url "https://raw.githubusercontent.com/jorgebucaran/fisher/master/fisher.fish"
     echo "fetching $url" >&2
+    command curl -s "$url?nocache" >$file.
 
-    set -l http_code (command curl -sw '%{http_code}' "$url?nocache" -o $file.)
-    set -l curl_code $status
-
-    if test $http_code -ne 200
-        echo -n "fisher: cannot update fisher -- " >&2
-
-        if test $curl_code -eq 0
-            echo "got http code $http_code, expected 200" >&2
-        else
-            echo "curl exited with code $curl_code, expected 0" >&2
-        end
-        command rm -f $file.
-        return 1
-    end
-
-    set -l next_version (command awk '{ print $4 } { exit }' <$file.)
+    set -l next_version (command awk '$4 ~ /^[0-9]+\.[0-9]+\.[0-9]+$/ { print v=$4 } { exit !v }' <$file.)
     switch "$next_version"
-        case $fisher_version
+        case "" $fisher_version
             command rm -f $file.
+            if test -z "$next_version"
+                echo "fisher: cannot update fisher -- are you offline?" >&2
+                return 1
+            end
             echo "fisher is already up-to-date" >&2
         case \*
             echo "linking $file" | command sed "s|$HOME|~|" >&2
--- a/fisher.fish
+++ b/fisher.fish
@@ -156,20 +156,25 @@ function _fisher_self_update -a file
     set -l url "https://raw.githubusercontent.com/jorgebucaran/fisher/master/fisher.fish"
     echo "fetching $url" >&2
 
-    if not set -l curl_response (command curl -fsS "$url?nocache" -o $file. 2>&1)
-        echo "fisher: cannot update fisher -- $curl_response" >&2
+    set -l http_code (command curl -sw '%{http_code}' "$url?nocache" -o $file.)
+    set -l curl_code $status
+
+    if test $http_code -ne 200
+        echo -n "fisher: cannot update fisher -- " >&2
+
+        if test $curl_code -eq 0
+            echo "got http code $http_code, expected 200" >&2
+        else
+            echo "curl exited with code $curl_code, expected 0" >&2
+        end
         command rm -f $file.
         return 1
     end
 
     set -l next_version (command awk '{ print $4 } { exit }' <$file.)
     switch "$next_version"
-        case "" $fisher_version
+        case $fisher_version
             command rm -f $file.
-            if test -z "$next_version"
-                echo "fisher: cannot update fisher -- unable to determine remote version" >&2
-                return 1
-            end
             echo "fisher is already up-to-date" >&2
         case \*
             echo "linking $file" | command sed "s|$HOME|~|" >&2
--- a/fisher.fish
+++ b/fisher.fish
@@ -155,14 +155,20 @@ end
 function _fisher_self_update -a file
     set -l url "https://raw.githubusercontent.com/jorgebucaran/fisher/master/fisher.fish"
     echo "fetching $url" >&2
-    command curl -s "$url?nocache" >$file.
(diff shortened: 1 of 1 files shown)


fisher.fish
```diff
@@ -155,14 +155,20 @@ end
...
     set -l url "https://raw.githubusercontent.com/jorgebucaran/fisher/master/fisher.fish"
     echo "fetching $url" >&2
-    command curl -s "$url?nocache" >$file.
+    set -l curl_response (command curl -fsS "$url?nocache" -o $file. 2>&1)
+
+    if test $status -ne 0
+        echo "fisher: cannot update fisher -- $curl_response" >&2
+        command rm -f $file.
+        return 1
+    end
```
jorgebucaran, line 164: @guihkx I think we could simplify it to this:

```fish
if not set -l curl_response (command curl -fsS "$url?nocache" -o $file. 2>&1)
    echo "fisher: cannot update fisher -- $curl_response" >&2
    command rm -f $file.
    return 1
end
```

Could you check again please?

jorgebucaran:
> @guihkx By using curl's -f option, we improve the self-update mechanism by not overwriting the fisher script if the HTTP status code given to us by Github is higher than 400.

Why do we need to duplicate the error handling or where do we fail here?

https://github.com/jorgebucaran/fisher/blob/994506e5945c8f9bb274dec17ebbe47e0a50cdb1/fisher.fish#L162-L167

guihkx:
I really considered removing that check, but curl's `-f` option only fails on HTTP codes equal or higher than 400.

But what if something goes really wrong with Github servers and they give us a 3xx code? We're currently not using curl's `-L` option to follow the Location header? Or even a 200 OK response but with wrong response body (e.g. empty or with some generic error)?

In this case I think it's better be safe than sorry. Also, that's an useful check in case `awk` fails (for whatever reason).

Thoughts?

jorgebucaran:
@guihkx Gotcha. Could there be a way to get whatever curl gives us, and use awk to validate the file avoiding handling errors twice? If you could list up any of the possible error responses, I'm happy to provide the awk code to do this.

fisher.fish
guihkx: Done! Thanks for the tip.

guihkx:
I just learned that we can use curl to give us the HTTP status code directly to `stdout`. For instance:

```
$ curl -fsw '%{http_code}' 'https://raw.githubusercontent.com/jorgebucaran/fisher/master/fisher.fish' -o $file.
```

If the output is different than `200`, we simply abort and show a generic error with the HTTP status code we got from Github.

I mean, I think it's safe to assume that if we get a HTTP 200 OK code, the content is "trustworthy".

Then that second check could be fully removed.

jorgebucaran:
@guihkx Interesting, where would that leave us?

guihkx:
I just pushed the modifications I did.

It only does a single check now: unless we get a 200 OK, the self-update will abort with an error:

image

I personally think the changes are sane, but let me know if you think it's "too much".

By the way, I'm not _that_ familiar with the `fish` syntax, so any optimizations you can think of are very much welcome! :)

jorgebucaran:
@guihkx Whoops, I think I miscommunicated here. I'm trying to both eliminate duplication (https://github.com/jorgebucaran/fisher/pull/580#issuecomment-664192976) and keep the code simple. Now, there's even more code. 😅 

If possible, I'd like to dump whatever curl returns in a temporary file, then use awk to validate the content.

guihkx:
Ohhhh, gotcha now! 😆 

But when you say "validate the content", are you intending to check if output given by curl is an _actual_, _valid_, fish script? If that's the case, I think it sounds too complicated and I have no idea how can we do that properly. Maybe if fish has a tool for validating the syntax of their scripts or something like that...?

jorgebucaran:
It has something "close", have you looked at `fish_indent`? But I don't mean let's use that or try anything too complicated. 

What we are doing right now is grabbing the version number from the file with awk. Do you know how that awk script doesn't fail even when curl does? I understand that's part of the problem we have here.

guihkx:
>It has something "close", have you looked at `fish_indent`?

I took a quick look at the documentations right now, but I'm no that sure it would be useful... However, the fish manual pointed me to the `-n` flag:

```
       • -n or --no-execute do not execute any commands, only perform syntax checking
```

It seems pretty fast, too:

```
$ time fish -n ~/.config/fish/functions/fisher.fish &>/dev/null

________________________________________________________
Executed in    9,51 millis    fish           external 
   usr time    6,81 millis  276,00 micros    6,54 millis 
   sys time    2,27 millis   84,00 micros    2,19 millis 

$ echo $status
0
```

Testing with a random, non-script file (500 KiB):

```
$ time fish -n test.svg &>/dev/null

________________________________________________________
Executed in   24,37 millis    fish           external 
   usr time   18,34 millis  396,00 micros   17,94 millis 
   sys time    6,04 millis   58,00 micros    5,98 millis
```

>Do you know how that awk script doesn't fail even when curl does? I understand that's part of the problem we have here.

I think I do. As I mentioned in #578, Github went through a power outage and this is the raw response their server was giving me:

```
500: Internal Server Error
```

Running that output on `awk`:

```
$ echo '500: Internal Server Error' | awk '{ print $4 } { exit }'
Error
```

In order for the self-update to fail, the output of the command above would have to be an empty string, right?

jorgebucaran:
@guihkx Hmm, then this might just be everything we need:

```fish
set -l next_version (command awk '$4 ~ /^[0-9]\.[0-9]\.[0-9]/ { print v=$4 } { exit !v }' <$file.)
```

Could you test it?

guihkx:
I was just about to suggest using regex, but I had no idea on how to do that with awk. x)

Does fisher follow [semantic versioning](https://semver.org/)? For instance that check would fail if you released version 3.10.1.

jorgebucaran:
@guihkx Yes, whoops, my bad. This should work:
```awk
awk '$4 ~ /^[0-9]+\.[0-9]+\.[0-9]+$/ { print v=$4 } { exit !v }'
```

guihkx:
Just tested it, looks good!

Don't forget to squash this whole mess before merging. 😅

Thanks a bunch!

jorgebucaran:
Thank you, @guihkx ! [`3.2.12`](https://github.com/jorgebucaran/fisher/releases/tag/3.2.12) is now officially out! All fishfolk rejoice! 🦐

PR #597, opened 2020-11-04
Use fish_plugins filename instead of old fishfile.

aureooms:


jorgebucaran:
Whoops! Thank you, @aureooms!

PR #627, opened 2020-11-23
Use HTTPS for installation

rany2:
Using HTTPS would help prevent: 

* accidentally running a captive portal login screen as a fish script and sourcing it
* and other obvious security and privacy benefits of HTTPS

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Looking for plugins? Browse [git.io/awsm.fish](https://git.io/awesome.fish) or [
 ## Installation
 
 ```console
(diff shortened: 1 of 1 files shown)


jorgebucaran:
> accidentally running a captive portal login screen

How could this happen if GitHub owns Git.io?

rany2:
You know when you're connected to hotel WiFi and you need to login to the WiFi network? If it's over http then it could load the captive login screen whereas with https this isn't possible.

jorgebucaran:
Got it! Thanks and sorry for the delay here.

Issue #635, opened 2020-12-11
Support comments in fish_plugins file


mattmc3:
It would be very helpful to be able to comment out a plugin, but not remove it from the `fish_plugins` file. Are you open to skipping over lines that begin with a hash (#) as a comment?

jorgebucaran:
@mattmc3 Could you elaborate on your use case a bit more?

jorgebucaran:
Could you elaborate on your use case a bit more?

mattmc3:
I can see a few different important use cases.

Some plugin names aren't really descriptive about what they are or what they do. For example, edc/bass is cleverly named, but not obvious what it's for, so it's nice to have something like this in fish_plugins:

```shell
# bass is used to make it easy to use utilities written for Bash in fish
edc/bass
```

Secondly, as a fish plugin developer, it's also handy to comment out my own published fish plugins when I'm making changes.

```shell
# Normally I want my fish-up plugin, but as I make changes to it it's nice to comment it out
# and use my dev version in a different place on my machine.
# mattmc3/fish-up
```

And finally, sometimes I try out a plugin, but then remove it but want to keep it around for reference.

```shell
# plugins I use
jorgebucaran/fisher
mattmc3/fish-up

# archived plugins
# edc/bass
```

This is not a novel concept, as many other plugin managers do this. For example, a few popular ZSH ones do this:
- [Antibody](https://getantibody.github.io/usage/)
- [ZGen](https://github.com/tarjoilija/zgen/blob/0b669d2d0dcf788b4c81a7a30b4fa41dfbf7d1a7/zgen.zsh#L398)

I can understand that fisher is a little different in that it modifies the fish_plugins file rather than having it all be hand edited, so doing things like `fisher remove` might leave comment artifacts, but that's totally fine. If you are hand editing your fish_plugins file to add comments, that's the price you pay. Hope this helps see the usefulness.

jorgebucaran:
@mattmc3 Take it for a spin:

```fish
fisher remove jorgebucaran/fisher
fisher install jorgebucaran/fisher@comments
```

> For example, edc/bass is cleverly named, but not obvious what it's for...

Maybe [replay.fish](https://github.com/jorgebucaran/replay.fish) has the best of both worlds. 😉

jorgebucaran:
[...]

PR #653, opened 2021-01-18
Implements comments for fish_plugins

soraxas:
This PR aims to implement a simple comment system to address #635.

Commit 11eb085 had linked #524 of deprecating `fishfile` for `fish_plugins`, which is nice but comments supports are dropped, perheps due to the previous `filefile`'s complexity.

This PR aims to introduce a simple and easy to maintain comment system back to `fish_plugins` (partly based on ae770fc). Currently, this commit works on both full line comments and inline comments (where there should be at least one whitespace before the first # character), using only built-in fish-shell string regex.

## Explaination:
The following
```fish
string match --regex -- '^[^#\s]+' <$fish_plugins
```
can essentially strip away any full or inline comments as it captures from the beginning of the line to the first non-whitespace non-# character.

Then, in the logic of writing back to `$fish_plugins`, we first use
```fish
# match line comments, active plugins (+ optional inline comments) in fish_plugins
set --local -- new_content (string match --regex -- '^\s*#.*$|^(?:'(
    string join -- '|' $_fisher_plugins)')(?:\s+#.*)?$' <$fish_plugins)
```
to construct a regex with the installed active plugins to filter out the content of `$fish_plugins`. The regex only match fullline comments (`'^\s*#.*$`), active plugins, and an optional inline-comment capture group (`(?:\s+#.*)?`) after the active plugins. Note that we had checked `$_fisher_plugins[1]` to be defined so the string join should always has at least one item.

Finally, 
```fish
# inverse match active plugins against the (comment stripped) $new_content
set --append -- new_content (string match --regex --invert -- '^(?:'(
    string join -- '|' (string match --regex -- '^[^#\s]+' $new_content))')$' $_fisher_plugins)
```
first strips away any full/inline comments from `$new_content`, then performs an exact inverse match so that any newly installed plugins in `$_fisher_plugins` that are currently not present in `$fish_plugins` would be added to `$new_content`. As an added bonus, the contents and comments in `$fish_plugins` will be order-preserved with this approach, and newly installed plugins are appended at the end of the file.

### DIFF
--- a/.github/workflows/fork-sync.yaml
+++ b/.github/workflows/fork-sync.yaml
@@ -1,0 +1,20 @@
+name: Sync Fork
+
+on:
+  schedule:
+    - cron: '*/30 * * * *' # every 30 minutes
+  workflow_dispatch: # on button click
+
+jobs:
+  sync:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: tgymnich/fork-sync@v1.3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          owner: jorgebucaran
+          base: main
+          head: main
+          merge_method: rebase
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -33,7 +33,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                 else if test ! -e $fish_plugins
                     echo "fisher: \"$fish_plugins\" file not found: \"$cmd\"" >&2 && return 1
                 end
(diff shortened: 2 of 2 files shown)


jorgebucaran:
Thank you for creating a new POC, but I just closed #635 stating that I will not be adding comment support for now. 

BTW, I think you forgot to [`string escape`](https://fishshell.com/docs/current/cmds/string-escape.html) in your implementation to encode plugin names.

jorgebucaran:
@soraxas Renabling support for top-level `.fish` files in this PR was uncalled for. Please take a break.

PR #661, opened 2021-02-20
Add basic manual page support

marcransome:
@jorgebucaran I've been toying with the ability to install man pages with `fisher` plugins and would like your input if possible. The changes I've proposed here add basic functionality for man pages, and can be seen to work when using the [pond](https://github.com/marcransome/pond) plugin:

```console
$ fisher install marcransome/pond@feature/stdin
fisher install version 4.2.0
Fetching https://codeload.github.com/marcransome/pond/tar.gz/feature/stdin
Installing marcransome/pond@feature/stdin
           /home/nobody/.config/fish/functions/pond.fish
           /home/nobody/.config/fish/conf.d/pond.fish
           /home/nobody/.config/fish/completions/pond.fish
           /home/nobody/.config/fish/manpages/pond.1
Installed 1 plugin/s
```

The man page should be available as expected:
```
$ man pond
...
```

<hr>

Files matching the pattern `manpages/<name>.<section-number>` are installed to `$fisher_path/manpages` and symbolic links are created at `/usr/share/man/man<section-number>/<name>.<section-number>`. Manual files and symbolic links are removed when a plugin is uninstalled.

I'd like to get some feedback regarding the changes, as well as any thoughts you might have on how to handle different locales (e.g. `<my-project>/manpages/<locale>/<name>.<section-number>`. 🤔

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -186,11 +186,11 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                 end
 
                 for man in (string match --regex -- '.+/manpages/([^/]+)\.(\d)$' $$plugin_files_var)
-		            if set --local name (string match --regex -- '.+/([^/]+)\.(\d)' $man)
+                    if set --local name (string match --regex -- '.+/([^/]+)\.(\d)' $man)
                         if test (ln -s $fisher_path/manpages/$name[2].$name[3] /usr/share/man/man$name[3]/$name[2].$name[3]) $status -ne 0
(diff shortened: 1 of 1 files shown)


jorgebucaran:
I'm cautious of introducing Fisher-only features. My intention is not really to extend out-of-the-box Fish, but to make it easier to install, update, and remove "plugins", which essentially inherit their look and feel from `~/.config/fish`.

For some history, we used to have `man` page support in 1.x~2.x, but I ended up removing it. AFAIK `~/.config/fish/manpages` is not a special place for users to deposit general-purpose man pages that Fish knows about.

I'd like to hear what you think. Maybe we can agree to support this in a different form, perhaps even through a third-party plugin.

jorgebucaran:
I'm still interested in having that discussion, but I haven't received a reply in a while, so I'm just going to close for now.

PR #662, opened 2021-02-24
Make "search on GitHub" links more obvious

shuyuan-liu:
I placed the links to the three GitHub topics (fish-plugin, fish-plugins and fish-package) after "search on GitHub". The original version where each of "search", "on", and "GitHub" links to one of those topics seemed a bit confusing to me.


jorgebucaran:
Thanks, but no thanks! I fully agree that your changes make it less confusing, but I really prefer the way it looks right now.

Issue #489, opened 2018-11-05
Plugin installation fails on OpenBSD


bertalanp99:
I managed to install `fisher` on my OpenBSD box by downloading the function w/ `curl` as recommended. `fisher` itself works, but this happens when I try installing any plugin:

    $ fisher add jethrokuan/z
    fetching https://codeload.github.com/jethrokuan/z/tar.gz/master
    cannot install github.com/jethrokuan/z -- is this a valid package?
    done in 1.00s

Thank you for your help in advance!

jorgebucaran:
@bertalanp99 How can I get my hands on an OpenBSD box? Could you help with that?

jorgebucaran:
@bertalanp99 In the meantime, can you try a few things for me?

First, I want to make sure we can get the tarball. The following should write it to standard output. It should look like a lot of nonsense.
```fish
curl -Ss https://codeload.github.com/jethrokuan/z/tar.gz/master 2>&1 
```

If that works, let's see if we can extract the package. The following should extract the package to a directory named `tmp` in your current working directory.

```fish
mkdir tmp
curl -Ss https://codeload.github.com/jethrokuan/z/tar.gz/master 2>&1 |  tar -xzf- -C ./tmp --strip-components=1
la tmp
```

jorgebucaran:
@bertalanp99 I meant how to virtualize it. So, now we know the problem is `tar`.

jorgebucaran:
@bertalanp99 I meant how to virtualize it. I managed to virtualize FreeBSD, not OpenBSD (which does support `--strip-components` by the way).

The problem must be `tar` then. We're not actually writing the package to the terminal, so we can leave `curl` as it is.

bertalanp99:
@jorgebucaran Oh, I see. I don't really know, I haven't ever done it myself, but maybe I can look into it later today. You mean the problem could be because of `--strip-components` in `tar`?

jorgebucaran:
@bertalanp99 The problem is definitely that OpenBSD's `tar` doesn't support `--strip-components` as you said. The solution is to extract the package separately. It's a bit more involved and less elegant than using the pipeline as it is currently implemented, but if that's the price we have to pay to support OpenBSD, so be it. 

I'll ping you when I'm done implementing it.

jorgebucaran:
@bertalanp99 What git version are you using?

bertalanp99:
I am using `v2.19.1`, why?

jorgebucaran:
[...]

PR #665, opened 2021-03-10
fisher.fish: Use short-option style for tar

jagtalon:
OpenBSD's tar command unfortunately doesn't support long options (see https://man.openbsd.org/tar). Using short options ensures compatibility with OpenBSD systems.

Also see https://github.com/jorgebucaran/fisher/issues/489

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -86,7 +86,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
 
-                        if curl --silent \$url | tar --extract --gzip --directory \$temp --file - 2>/dev/null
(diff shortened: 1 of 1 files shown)


functions/fisher.fish
```diff
@@ -86,7 +86,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
 
-                        if curl --silent \$url | tar --extract --gzip --directory \$temp --file - 2>/dev/null
+                        if curl --silent \$url | tar -xz -C \$temp -f - 2>/dev/null
```
jorgebucaran, line 89: I suppose `-xzC` works too? Or do you find -xz -C preferable for some reason? 😏

jorgebucaran:
Thank you, @jagtalon! 💯

PR #685, opened 2021-11-05
prevent error on unzipping packages

mehdipourfar:
I got this error on installing fisher:
`
fisher install version 4.3.0
Fetching https://codeload.github.com/jorgebucaran/fisher/tar.gz/HEAD
fisher: Invalid plugin name or host unavailable: "jorgebucaran/fisher"
`
and I've found out that piping the results of curl into tar will causes the problem. This commit fixes that.

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -81,18 +81,20 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                         command cp -Rf $plugin/* $source
                     else
                         set temp (command mktemp -d)
+                        set dlpath (command mktemp)
                         set name (string split \@ $plugin) || set name[2] HEAD
                         set url https://codeload.github.com/\$name[1]/tar.gz/\$name[2]
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Hi, @mehdipourfar, thank you! What's your OS and Fish version?

mehdipourfar:
Hi
```
lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal
```

```
fish --version
fish, version 3.3.1
```

jorgebucaran:
Could you perhaps prepare an image with the same OS, tar, and curl version where we can reproduce this issue? I don't really want to merge something until I understand for whom Fisher is currently broken.

mehdipourfar:
Finally I've found out the problem. I don't know when and why I've had added this in my fish functions:
```fish
function curl
    command curl -w "\n" $argv
end
```

jorgebucaran:
Do you have a curlrc file with a custom curl configuration?

Issue #670, opened 2021-03-28
fisher update returns conflicting files errors and removes fish_plugins file


jimblue:
Hello,

I'm having some issues with fisher update command, when I run it I get the following errors :

```
fisher update
fisher update version 4.3.0
Fetching https://codeload.github.com/jorgebucaran/fisher/tar.gz/HEAD
Fetching https://codeload.github.com/adriaanzon/fish-artisan-completion/tar.gz/HEAD
Fetching https://codeload.github.com/dracula/fish/tar.gz/HEAD
Fetching https://codeload.github.com/jethrokuan/fzf/tar.gz/HEAD
Fetching https://codeload.github.com/jethrokuan/z/tar.gz/HEAD
fisher: Cannot install "jorgebucaran/fisher": please remove or move conflicting files first:
        /Users/jimblue/.config/fish/functions/fisher.fish
fisher: Cannot install "adriaanzon/fish-artisan-completion": please remove or move conflicting files first:
        /Users/jimblue/.config/fish/functions/artisan.fish
        /Users/jimblue/.config/fish/completions/artisan.fish
        /Users/jimblue/.config/fish/completions/php.fish
fisher: Cannot install "dracula/fish": please remove or move conflicting files first:
        /Users/jimblue/.config/fish/conf.d/dracula.fish
fisher: Cannot install "jethrokuan/fzf": please remove or move conflicting files first:
        /Users/jimblue/.config/fish/functions/__fzf_cd.fish
        /Users/jimblue/.config/fish/functions/__fzf_complete.fish
        /Users/jimblue/.config/fish/functions/__fzf_complete_preview.fish
        /Users/jimblue/.config/fish/functions/__fzf_find_file.fish
        /Users/jimblue/.config/fish/functions/__fzf_get_dir.fish
        /Users/jimblue/.config/fish/functions/__fzf_open.fish
        /Users/jimblue/.config/fish/functions/__fzf_parse_commandline.fish
        /Users/jimblue/.config/fish/functions/__fzf_reverse_isearch.fish
        /Users/jimblue/.config/fish/functions/__fzfcmd.fish
        /Users/jimblue/.config/fish/conf.d/fzf.fish
fisher: Cannot install "jethrokuan/z": please remove or move conflicting files first:
        /Users/jimblue/.config/fish/functions/__z.fish
        /Users/jimblue/.config/fish/functions/__z_add.fish
        /Users/jimblue/.config/fish/functions/__z_clean.fish
[...]

Issue #644, opened 2021-01-07
Unending loop when trying to automatically install fisher.


Jomik:
I had an auto install snippet in my `config.fish`
```fish
if not functions -q fisher
    curl https://git.io/fisher --create-dirs -sLo $XDG_CONFIG_HOME/fish/functions/fisher.fish
    fish -c fisher
end
```
Trying to use the install method, breaks, as it will never see that `fisher` exists in functions.
```fish
if not functions -q fisher
    curl -sL https://git.io/fisher | source && fisher install jorgebucaran/fisher
end
```
Is there a nice way to automate this?

jorgebucaran:
Do it only if you're interactive.

```fish
if status is-interactive && ! functions --query fisher
    curl --silent --location https://git.io/fisher | source && fisher install jorgebucaran/fisher
end
```

Jomik:
Of course! Thank you 👍

PR #686, opened 2021-11-13
Docs: add bootstrap note 

luong-komorebi:
This note is important because there are certain caveats when bootstrapping fisher:
1. User without this instruction might run into a loop. #644
2. User who follows the standard installation process and trying to couple that into a bootstrap script will run into this issue #670 where the whole fish_plugins is overwritten

I borrowed the wording from this commit  dadcc6c

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,19 @@ Manage functions, completions, bindings, and snippets from the command line. Ext
 ## Installation
 
 ```console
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Hey, @luong-komorebi, thanks, but I'll pass. We don't promote bootstrapping Fisher anywhere like that, precisely because of all the caveats, so while this comment is useful, we don't need it at this time.

PR #700, opened 2022-01-20
updated github api endpoint

### LINKED ISSUE #699: fisher uses a seemingly undocumented github api endpoint
Fisher makes a request to `codeload.github.com` to retrieve the tarball of the desired plugin. Upon research(basically googling for `+"codeload.github.com" +site:docs.github.com`), I couldn't find proper documentation for this endpoint.

As there are no guarantees this will work indefinitely or that I'll have consistent behavior, I suggest moving to the well-documented rest api. Then I'd also feel more comfortable with advanced usage of the api, such as the ability to specify a "tag, branch, or commit". 

Relevant section: https://docs.github.com/en/rest/reference/repos#download-a-repository-archive-tar

anakimluke:
We need the -L flag because we're being redirected.
The included header is recommended by github.

closes #699

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -84,11 +84,12 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                     else
                         set temp (command mktemp -d)
                         set name (string split \@ $plugin) || set name[2] HEAD
-                        set url https://codeload.github.com/\$name[1]/tar.gz/\$name[2]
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Is this enough to close #699?

anakimluke:
yep I think so

halostatue:
I just made the changes here to my local copy of `fisher.fish`, and it worked perfectly.

PR #704, opened 2022-02-03
Improve fisher autocompletions

### LINKED ISSUE #703: Autocompletions don't work when passing variables on the command line
Autocompletions don't work when, (for example) manually specifying `$fisher_path`:
```
fisher_path=/some/nice/path fisher [TAB]
```
does not offer the usual options.

This should be easy to fix: the [fish completion tutorial](https://fishshell.com/docs/current/completions.html) seems to suggest something like `set --local fisher_subcommands install update remove list` and then using
```
not __fish_seen_subcommand_from $fisher_subcommands
```
as the condition, instead of `__fish_use_subcommand`.

However, I am inclined to see this as an upstream issue with [`__fish_use_subcommand`](https://github.com/fish-shell/fish-shell/blob/master/share/functions/__fish_use_subcommand.fish).

alexrutar:
Fix autocompletions when called like `fisher_path=/some/path fisher [TAB]`.

Closes #703

### DIFF
--- a/completions/fisher.fish
+++ b/completions/fisher.fish
@@ -1,7 +1,8 @@
+set --local fisher_subcommands install update remove list
 complete --command fisher --exclusive --long help --description "Print help"
 complete --command fisher --exclusive --long version --description "Print version"
-complete --command fisher --exclusive --condition __fish_use_subcommand --arguments install --description "Install plugins"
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Closing as per https://github.com/jorgebucaran/fisher/issues/703#issuecomment-1097363226.

PR #710, opened 2022-04-13
Updated README.md to copy the default install of tide

AucaCoyan:
I changed the tide install for the current install of tide in [their readme](https://github.com/IlanCosman/tide):
before:
```console
fisher install ilancosman/tide
```
after
```console
fisher install IlanCosman/tide@v5
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ You can install, update, and remove plugins interactively with Fisher, taking ad
 Install plugins using the `install` command followed by the path to the repository on GitHub.
 
 ```console
-fisher install ilancosman/tide
+fisher install IlanCosman/tide@v5
 ```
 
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Care to explain why?

AucaCoyan:
Sorry for the delay! I wanted to fix the readme for updated installation command. 
I believe it is probable that the same user that installs fisher will also instal tide, so, to ease the path, I would copy and paste the same command of tide. Is that right?

jorgebucaran:
Thank you, @eitsupi , I think that's the piece of info I was missing.

jorgebucaran:
We'll need to move Tide to the second installation example, which is where we explain the use of @ to install tags or branches.

AucaCoyan:
Thanks! You are right, I didn't explained myself correctly. I will take this as a lesson! 😄 .
l feel satisfied to close the PR. Have a nice day!

PR #716, opened 2022-05-23
Use current fish path instead of fish command

IlanCosman:
`fish` may not be an available command, for example if it was installed through `brew`. Whenever I `rm -r ~/.config/fish/` on my mac (for developing Tide), I have to `fish_add_path /opt/homebrew/bin/` before I run my `fisher` command. Instead, lets just guarantee a valid command by using the path to the current instance of fish. Tide already does this and it has worked fine so far.

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -72,13 +72,13 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
             set --local fetch_plugins $update_plugins $install_plugins
             echo (set_color --bold)fisher $cmd version $fisher_version(set_color normal)
 
-            for plugin in $fetch_plugins
+            fish_path=(status fish-path) for plugin in $fetch_plugins
                 set --local source (command mktemp -d)
                 set --append source_plugins $source
 
                 command mkdir -p $source/{completions,conf.d,functions}
 
-                fish --command "
+                $fish_path --command "
                     if test -e $plugin
                         command cp -Rf $plugin/* $source
                     else


jorgebucaran:
@IlanCosman This requires at least what version of Fish?

IlanCosman:
Do you mean when did `status fish-path` get released? It's been there since Fish 3.0 https://github.com/fish-shell/fish-shell/commit/e212269ab1b29a1456ce47c335ae57fb5789ea79. 

And I'm pretty sure `foo=bar` as a way of setting variables has been around for a while as well. Can always change it to `set` if not.

functions/fisher.fish
```diff
@@ -72,13 +72,13 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
...
 
-            for plugin in $fetch_plugins
+            fish_path=(status fish-path) for plugin in $fetch_plugins
                 set --local source (command mktemp -d)
                 set --append source_plugins $source
 
                 command mkdir -p $source/{completions,conf.d,functions}
 
-                fish --command "
+                $fish_path --command "
```
jorgebucaran, line 81: Is this also Fish 3.0 or was it introduced more recently?

jorgebucaran: I know this has been around for a while now. Just checking.

IlanCosman: @faho Sorry to bother but would you happen to know when variable-in-command-position was released? Tried combing through the release notes but couldn't find it.

jorgebucaran: It seems that this is in since 3.0 as well! 🎉

jorgebucaran: https://github.com/fish-shell/fish-shell/commit/865a4647aecbd7bf0fe55cb26f176a6646d4aed0

jorgebucaran (approved):
LGTM

Will merge here and release tomorrow! Thanks.

jorgebucaran:
> fish may not be an available command, for example if it was installed through `brew`. 

Doesn't brew put Fish in your PATH? I install Fish via brew too and I've never had to do that. Or I could've added it myself and forgotten about it. 😅

IlanCosman:
> Doesn't brew put Fish in your PATH?

`/opt/homebrew/bin/` isn't in PATH by default. Has to get added, If you want to test it, back up your `~/.config/fish` and then delete it. You shouldn't have access to any commands installed through `brew`.

jorgebucaran:
Yes. Then I must have added it to my PATH. Makes sense. 👍

PR #719, opened 2022-06-17
Fix a compatibility issue with fish 3.0.x

halostatue:
I have a docker image that can install fish, but it is verison 3.0.2. The bash-compatible mechnism of `fish_path=(status fish-path) command` was not added until after 3.0.2, probably 3.1. By setting `fish_path` *earlier* in the fuction, Fisher has full compatibility with this earlier version of Fish 3.


jorgebucaran (approved):

jorgebucaran:
Thank you, @halostatue, perfectly done. 🙌

PR #735, opened 2022-08-29
added --insecure flag to curl

rolandweibel:
Added --insecure flag to curl to fix error when certificates were not trusted

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -98,7 +98,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
 
-                        if curl --silent -L \$url | tar -xzC \$temp -f - 2>/dev/null
+                        if curl --insecure --silent -L \$url | tar -xzC \$temp -f - 2>/dev/null
                             command cp -Rf \$temp/*/* $source
                         else
                             echo fisher: Invalid plugin name or host unavailable: \\\"$plugin\\\" >&2


jorgebucaran:
@rolandweibel Why were certificates not trusted?

jorgebucaran (approved):
👍

jorgebucaran:
Got it, thanks, and this flag basically tells curl to go ahead regardless.

I wonder if anyone more knowledgeable about curl could point out if there are any disadvantages to adding this. 🤓

waeltken:
I don't think this would be a good default, maybe as an option? I would not want fisher to pull from servers without valid TLS certificates since there is no other signature or integrity check for the plugins.

jorgebucaran:
Agreed, but Fisher won't pull from any servers other than GitHub or GitLab's, right?

2-4601:
With the `--insecure` flag `curl` cannot distinguish a real GitHub server from a fake one. I.e. without certificates a malicious actor can pretend to be GitHub and feed you a bad Fish theme that turns all your colors pink and cyan.

jorgebucaran:
> a malicious actor can pretend to be GitHub 

How would that actually work? You can't feed URLs to Fisher. `foo/bar` means install from GitHub (by default) or from GitLab by using `gitlab.com/` before the plugin name. 🤔 

> a bad Fish theme that turns all your colors pink and cyan.
So bad, indeed! 🙈

jorgebucaran:
> a malicious actor can pretend to be GitHub 

How would that actually work? You can't feed URLs to Fisher. `foo/bar` means install from GitHub (by default) or from GitLab by using `gitlab.com/` before the plugin name. 🤔 

> a bad Fish theme that turns all your colors pink and cyan.

So bad, indeed! 🙈

2-4601:
> How would that actually work?

Let's say you don't want to work from home or office today. So you go outside to a park or a cafe. You pick a Wi-Fi network and start working. But before you start you must procrastinate a little bit and you decide to install the Fish logo. So you run `fisher install laughedelic/fish_logo`. When you run this, `curl` will ask _"Who's `github.com`?"_. Let's assume that someone can send you a false reply or that the DNS server itself is poisoned or compromised and it sends you the wrong IP address. Now when `curl` establishes the connection to `github.com` with the `--insecure` flag, it will happily connect to the wrong IP without objections. And the fake GitHub can now serve `curl` anything it pleases.
However, if `curl` is run **without** the `--insecure` flag, it will see that the fake GitHub's certificate is not trusted, refuses to continue and throws an error.

functions/fisher.fish
```diff
@@ -98,7 +98,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
 
-                        if curl --silent -L \$url | tar -xzC \$temp -f - 2>/dev/null
+                        if curl --insecure --silent -L \$url | tar -xzC \$temp -f - 2>/dev/null
```
waeltken, line 101: This disables SSL certificate validation! 🛑

jorgebucaran:
Thank you, @rolandweibel and @2-4601. Closing as wontfix.

Issue #733, opened 2022-08-13
Add Codeberg support


aral:
Would love to be able to move my little plugin (https://github.com/small-tech/gills) to [Codeberg](https://codeberg.org) (as I’m migrating projects off of Github… for reasons.)

This is where you’d implement it: 

https://github.com/jorgebucaran/fisher/blob/d431f45abcd89d6149a6b894a0a98ba02afa3703/functions/fisher.fish#L92-L97

Related: #657

aral:
PS. Happy to work on this when I get a moment but opening an issue first in case there’s an objection and/or if anyone else wants to tackle it.

jorgebucaran:
Sorry for the late reply. Why do you think Fisher should support Codeberg?

aral:
@jorgebucaran Hj Jorge, no worries. I guess the question is more “Why do you think Fisher should not support Codeberg?” :)

The answer to the other question is in my original request: so that folks can have the option of hosting their plugins on an ethical, cooperative, not for profit git host instead of one run by a trillion-dollar US corporation that mines their code to feed their commercial AI product. 

Given that it’s so easy to add support for this, is giving people this option something you’d be opposed to doing? If so, I’d love to know why :)

jorgebucaran:
I just recently found out about Codeberg and I'm trying to decide if it's worth adding support for it. You made some good points.

aral:
Neat. Happy to hear it :)

jorgebucaran:
We have a PR for this topic here https://github.com/jorgebucaran/fisher/pull/737. Anyone interested can continue the discussion there. I'll be closing this now.

PR #737, opened 2022-09-06
✨ add codeberg.org support

waeltken:
Implements #733


jorgebucaran:
Sorry for leaving this unattended for so long. I will not move forward with this. For now, Fisher only supports GitHub and GitLab, which are the platforms I'm most familiar with. Maybe someday.

PR #740, opened 2022-10-08
Dereference links on install

carlfriedrich:
I have a plugin which shares code for zsh and fish. I'd like to symlink this shared code file to the `functions` directory within the git repository, so that it gets installed by fisher. Unfortunately fisher does not dereference links and hence installs a broken link in this case.

This PR adds the `-L` argument to `cp` in order to dereference links on install and update.

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -173,7 +173,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                 end
 
                 for file in (string replace -- $source/ "" $files)
-                    command cp -Rf $source/$file $fisher_path/$file
+                    command cp -RLf $source/$file $fisher_path/$file
                 end
 
                 set --local plugin_files_var _fisher_(string escape --style=var -- $plugin)_files


jorgebucaran:
Cool! What does dereferencing mean? Will this affect how we install other plugins?

carlfriedrich:
> Cool! What does dereferencing mean?

It means: "If the file to be copied is a symbolic link, do not copy the link, but the file it points to instead."

> Will this affect how we install other plugins?

It will affect symbolic links only. The installation of plugins containing regular files (which I assume are all existing plugins to date) won't change at all.

jorgebucaran:
Thanks!

jorgebucaran:
Thanks! Do you know if -L is portable alright? What OS did you try this on?

jorgebucaran:
Thanks! Do you know if -L is portable alright?

carlfriedrich:
At least it's part of GNU coreutils [since 22 years](https://github.com/coreutils/coreutils/commit/c93a607f058662ed2ab1984a381e07a85ddef2c9) and busybox [since 14 years](https://github.com/brgl/busybox/commit/c0431ed455a601ddbfcd9f30d6397d4e9145cf54), so I would assume so. Do you know if there are any other `cp` variants?

jorgebucaran:
@carlfriedrich On macOS it reads:      

```console
-L    If the -R option is specified, all symbolic links are followed.
```

carlfriedrich:
> @carlfriedrich On macOS it reads:
> 
> ```
> -L    If the -R option is specified, all symbolic links are followed.
> ```

Great, that's exactly what it's supposed to do.

carlfriedrich:
@jorgebucaran So will you merge this or are there any doubts left?

jorgebucaran:
Yes, of course! Sorry to keep you waiting and thanks!  🙌

carlfriedrich:
@jorgebucaran Great, thanks a lot! ❤️ Do you have plans for an upcoming release containing the patch?

jorgebucaran:
You bet! Should be good now.

carlfriedrich:
Yaaaay! 🥳 Thanks a lot for your quick support!

PR #745, opened 2022-10-28
Add support for private or protected fish plugins

alexmarnell:
## Context

I have a specific use case where a set of fish plugins are made available internally. They are not public Github repositories. The `curl` command used by fisher will only fetch plugins from public repositories. 

## Proposed Changes

By changing the command to `curl -n` it can take advantage of any credentials stored in the `~/.netrc` file. This may not be the best flag to have enabled by default, or perhaps goes against the spirit of sharing fish plugins publicly. An alternative could be to move this behind a variable or some other indicator to mark that a plugin is private.

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -98,7 +98,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
 
-                        if curl --silent -L \$url | tar -xzC \$temp -f - 2>/dev/null
+                        if curl --silent -L -n \$url | tar -xzC \$temp -f - 2>/dev/null
                             command cp -Rf \$temp/*/* $source
                         else
                             echo fisher: Invalid plugin name or host unavailable: \\\"$plugin\\\" >&2


jorgebucaran (changes requested):
Should include documentation.

jorgebucaran:
Would this be unsafe? We don't want to open a can of worms.

jorgebucaran:
Would this be unsafe? We don't want to open a can of worms.

jorgebucaran:
Would this be unsafe? We don't want to open a can of worms.

jorgebucaran:
@voidlock Hi, it's been some time. I'm generally fine with this minor change, but I'm just slightly worried it might pose a risk or harm to users. Thoughts?

voidlock:
@jorgebucaran I understand your concern. What if I updated the PR to include a mechanism similar to how Golang handles private modules with the `GOPRIVATE` environment variable? 

Perhaps a `FISHER_PRIVATE` environment variable that a user can customize.

```
set FISHER_PRIVATE github.com/your_github_username/mysecret github.com/your_github_username/myothersecret
```

When fetching plugins, Fisher would scan the list provided by `FISHER_PRIVATE`. If it's found, it can update the curl command to include the `--netrc` flag.

jorgebucaran:
Sorry for letting this sit for so long. That approach could potentially work, and I'm interested to see what you come up with first. I should've mentioned that earlier. 🙏

jorgebucaran:
A cheaper workaround is to enable netrc globally by adding it to your .curlrc:

```fish
echo netrc >> ~/.curlrc
```

Issue #611, opened 2020-11-10
Fisher should stop using universal variables for plugin state


lilyball:
Fisher 4.1.0 uses universal variables to maintain both the list of what plugins are installed, and what paths are installed for each plugin. This is potentially problematic in the scenario where I'm syncing my fish config between two machines, as having the universal variables get out of date may cause fisher to orphan installed files. There are two main scenarios here:

1. If I sync my fish config including `fish_variables` file. The fish documentation says not to modify the `fish_variables` file externally but to always use fish commands to do so. Experimentally, if I modify my `fish_variables` and then query variables from another pre-existing shell, the shell won't pick up on the variables until I've shown a new prompt first. This means if I sync my config and then immediately run `fisher update`, it will be using the old values of the universal variables, meaning if the list of files installed by plugins changed as part of the sync, fisher will then potentially orphan files.

2. If I sync my fish config but omit the `fish_variables` file (which represents what I actually do on my own machines), syncing changes to the list of installed plugin files will allow fisher to orphan files. Basically the same scenario as the first one except I don't have to run `fisher update` as the very first action after the sync, I can run it arbitrarily later instead.

Instead of using universal variables, fisher should put a file on disk in `$fisher_path` that describes what's currently installed. It's still possible for me to sync ignoring that file, but that's really my fault. Or it could even be clever and make the file something like `conf.d/_fisher_state.fish` which would set global variables, and fisher could re-source that file upon invocation (to ensure it picks up on any synced changes). Since any syncing I'm doing is presumably syncing that dir, this would ensure I don't skip the file by virtue of not allowlisting its path (which is to say, my original sync setup actually explicitly listed which paths inside `~/.config/fish` to sync instead of listing which files to ignore).

jorgebucaran:
@lilyball When you sync your fish config, do you sync everything inside `$__fish_config_dir` or just a few files?

jorgebucaran:
> @lilyball 2. [...] syncing changes to the list of installed plugin files will allow fisher to orphan files.

No, it will not, unless you modify Fisher's state in `$_fisher_plugins` and `$_fisher_PLUGIN_NAME_files`.

> @lilyball This is potentially problematic in the scenario where I'm syncing my fish config

[...]

PR #746, opened 2022-10-29
Initialize _fisher_plugins using fish_plugins file (#741)

### LINKED ISSUE #741: Committing plugin source files (completions, conf.d, functions, themes) is not supported
I would like to track not only the list of plugins (`fish_plugins` file) but also the source code of those plugins, for a few reasons. Typically when you run "fisher update" you don't know if anything changed in your plugins. Tracking the files in git makes it easier to review changes after you pull them in, and commit them to your dotfiles once you "approve" them. Secondly, it makes provisioning a new machine easier because you don't need to install Fisher manually, the code is already in your dotfiles.

The problem emerges when you want to run "fisher update" because the universal variables expected by Fisher are missing (I would prefer not to track the `fish_variables` file). Rather than suggest a duplicate of https://github.com/jorgebucaran/fisher/issues/611 ("Fisher should stop using universal variables for plugin state"), I wonder if it would be possible for Fisher to create the universal variables on startup if they don't exist?

```fish
if ! set --query _fisher_plugins
    # setup the Fisher universal variables by inspecting fish_plugins, etc.
end
```

Here is a minimal example: https://git.sr.ht/~crg/fishy

```
> docker container run --rm -it fishy
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
root@59bd29f2ca3c /# fisher update
fisher update version 4.4.3
Fetching https://api.github.com/repos/jorgebucaran/fisher/tarball/HEAD
Fetching https://api.github.com/repos/jorgebucaran/replay.fish/tarball/HEAD
fisher: Cannot install "jorgebucaran/fisher": please remove or move conflicting files first:
        /root/.config/fish/functions/fisher.fish
        /root/.config/fish/completions/fisher.fish
fisher: Cannot install "jorgebucaran/replay.fish": please remove or move conflicting files first:
        /root/.config/fish/functions/replay.fish
        /root/.config/fish/completions/replay.fish
```

charego:
This change is for users who commit fish_plugins and plugin sources but omit fish_variables.

On a system where Fisher's universal variables are unset, most Fisher commands do not work out-of-the-box:

* `fisher list [<regex>]`
  * returns nothing
* `fisher remove repo/plugin`
  * fails with error: "Plugin not installed: repo/plugin"
* `fisher update repo/plugin`
  * fails with error: "Plugin not installed: repo/plugin"
* `fisher update`
  * fails with error about conflicting files, deletes fish_plugins file
* `fisher install repo/plugin`
  * if fish_plugins contains the plugin, same behavior as bare `fisher update`
  * otherwise, installs the plugin, rewrites fish_plugins file with only "repo/plugin"

As of this commit all Fisher commands work except for `fisher remove`. Fisher still has no way of knowing which files to remove absent the universal variable that tracks the files associated to a plugin. It may make sense to reject calls like `fisher remove repo/plugin` if the `_fisher_repo_2F_plugin_files` universal variable is missing. Fisher could suggest the user run `fisher update` and try again.

Note: I haven't really investigated how this works with local plugins since I don't use those.

Attempt to fix https://github.com/jorgebucaran/fisher/issues/741.

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -2,6 +2,8 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
     set --query fisher_path || set --local fisher_path $__fish_config_dir
     set --local fisher_version 4.4.3
     set --local fish_plugins $__fish_config_dir/fish_plugins
+    test -e $fish_plugins && set --local file_plugins (string match --regex -- '^[^\s]+$' <$fish_plugins)
+    set --query _fisher_plugins || set --universal _fisher_plugins $file_plugins
 
     switch "$cmd"
         case -v --version
@@ -28,8 +30,6 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
             set --local arg_plugins $argv[2..-1]
             set --local old_plugins $_fisher_plugins
             set --local new_plugins
-
-            test -e $fish_plugins && set --local file_plugins (string match --regex -- '^[^\s]+$' <$fish_plugins)
 
             if ! set --query argv[2]
                 if test "$cmd" != update


charego:
Bump for further consideration, @jorgebucaran. Thanks.

jorgebucaran:
Thanks, @charego. I'll have a look at this as soon as I can manage. 👍

jorgebucaran:
I have been addressing and investigating issues all weekend, so I am taking a break for now. However, I wanted to let you know that I will certainly look into this matter soon and will incorporate this change (or a similar one, as I'm not certain this will be merged as is) before Fisher 5. This should occur in the not-too-distant future. 🙏

ZuBB:
is issue described here possibly related with one I have

```sh
vv@my-precious ~/.config/fish/conf.d $ fisher install oh-my-fish/plugin-bang-bang
fisher install version 4.4.3
Fetching https://api.github.com/repos/oh-my-fish/plugin-bang-bang/tarball/HEAD
fisher: Invalid plugin name or host unavailable: "oh-my-fish/plugin-bang-bang"
vv@my-precious ~/.config/fish/conf.d $ ll
```

jorgebucaran:
I doubt it. It might be related to the curl or tar version or your OS. What's your OS, distro, and the curl and tar version you're using?

https://github.com/jorgebucaran/fisher/blob/59228d6215cb4c36aff77008751736f540dc42d8/functions/fisher.fish#L101-L106

ZuBB:
> I doubt it. It might be related to the curl or tar version or your OS. What's your OS, distro, and the curl and tar version you're using?

its latest macos with m1 chip

```sh
vv@my-precious ~/work/own/careerpartner/src $ curl -V
curl 7.88.1 (x86_64-apple-darwin22.0) libcurl/7.88.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.51.0
Release-Date: 2023-02-20
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM NTLM_WB SPNEGO SSL threadsafe UnixSockets
vv@my-precious ~/work/own/careerpartner/src $ tar --version
bsdtar 3.5.3 - libarchive 3.5.3 zlib/1.2.11 liblzma/5.0.5 bz2lib/1.0.8 
vv@my-precious ~/work/own/careerpartner/src $ 
```

ZuBB:
okay, I figured it out 

https://github.com/jorgebucaran/fisher/pull/771

charego:
Essentially, how to use fisher with dotfiles? Is there a best practice as you see it? The instructions are geared toward interactive use.

jorgebucaran:
Honestly, the reason I haven't been able to work on this is because I don't understand the aim at a fundamental level. Do you just want Fisher to fail silently if universal variables are not set?

itsfarseen:
@jorgebucaran, from what I understand, the idea is to not use `fish_variables` to store fisher state as some of us don't commit `fish_variables` to our dotfiles repo.

So our options are:
* Reconstruct the state from the `fish_plugins` file and the list of installed files in `~/.config/fish/`;
* Provide a flag to `fisher update` that would let us overwrite the files present in `~/.config/fish`, but absent in `fish_variables`.

itsfarseen:
@jorgebucaran One more thing to consider before merging this, maybe we _should_ commit `fish_variables` to the dotfiles repo.

Plugins _should not_ store non-portable data in `fish_variables`.

For reference:
https://github.com/jorgebucaran/fisher/issues/611
https://github.com/jethrokuan/z/issues/104

jorgebucaran:
> @itsfarseen from what I understand, the idea is to not use fish_variables to store fisher state as some of us don't commit fish_variables to our dotfiles repo.

That sounds a lot like https://github.com/jorgebucaran/fisher/issues/611. Fisher preserves the '~' in paths when saving plugins to the universal state, making it safer to commit fish_variables. While it might not be perfect, it does prevent committing data that obviously varies across machines, such as different usernames or home directories. What else are we missing?

itsfarseen:
@jorgebucaran

> Fisher preserves the '~' in paths when saving plugins to the universal state

Yes, but Z doesn't. And possibly other plugins too.

OP's solution to this is to not commit `fish_variables` to the dotfiles repo.

My opinion is that, ideally we should ask Z and other problematic plugins to implement proper handling of `fish_variables`, instead of trying to make fisher work with missing `fish_variables`.

VinzNL:
Fish itself currently doesn't want you to share the fish_variables file across machines. For example, I share my dotfiles between two machines, one has fish 3.7.1 and one has fish 4.0.0. The first one has SETUVAR __fish_initialized:3400 and the second one has SETUVAR __fish_initialized:3800. There's even discussion about removing (default) uvars altogether, but that doesn't seem to be moving forwards quickly (fish-shell/fish-shell#7317 and fish-shell/fish-shell#7379). So we really need a way to sync plugins across machines without sharing fish_variables for the foreseeable future.

jorgebucaran:
Since [4.4.7](https://github.com/jorgebucaran/fisher/releases/tag/4.4.7), we've fixed one cause of fish_plugins becoming out of date during bootstrap via fisher install, which seems closely related to the main issue you were addressing here. I'm still monitoring other scenarios, but it would be helpful if you could check whether this situation is fixed for you now.

Issue #748, opened 2022-11-19
Autoinstall


Susensio:
How can I configure `~/.config/fish/config.fish` to autoinstall **fisher**, just like **fundle** does?

Fundle proposes adding `if not functions -q fundle; eval (curl -sfL https://git.io/fundle-install); end` to `config.d`. I've tried something similar with fisher with no luck.

Susensio:
I'm trying to figure out what's happening. Here goes a MRE:

conf.d/fisher.fish
```fish
if not functions -q fisher
    curl -sL https://git.io/fisher | source && fisher install jorgebucaran/fisher
end
```

and the output is
```
fisher install version 4.4.3
fisher install version 4.4.3
fisher install version 4.4.3
fisher install version 4.4.3
```

So it seem to be entering a infinite loop when calling `fisher install`, where it sources `conf.d/fisher.fish` again.

Susensio:
I found the problem: `fisher install` spawns `fish` subprocesses with `fish --command` that read configuration files, hence the infinite loop.
I've solved it checking if running from interactive session with `status is-interactive`.

This is the solution I've implemented (maybe this could be added to documentation):

```fish
if not functions -q fisher; and status is-interactive
    curl -sL https://git.io/fisher | source && fisher install jorgebucaran/fisher
end

I think this is a workaround and maybe `fisher install` should call `fish --command --no-config`, but I don't know for sure if this change could cause any problems.

jorgebucaran:
Close, but you'd still need to check for interactivity. So, that's the best solution we know of. 

You can use && instead of ; and in modern Fish, by the way.

jorgebucaran:
A PR to improve the docs is always welcome! 🙆‍♂️

PR #749, opened 2022-11-21
Improve docs and autoinstall behaviour

Susensio:
#748

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -82,7 +82,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                 command mkdir -p $source/{completions,conf.d,themes,functions}
 
-                $fish_path --command "
+                $fish_path --no-config --command "
                     if test -e $plugin
                         command cp -Rf $plugin/* $source
                     else
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@ Manage functions, completions, bindings, and snippets from the command line. Ext
 curl -sL https://git.io/fisher | source && fisher install jorgebucaran/fisher
 ```
 
+### Automatic install
+
+If you want to automatically install fisher when it is not present, you can add
+the following at the top of your `~/.config/fish/config.fish`.
+If `fish_plugins` is present is readed first.
+
+```fish
+if not functions -q fisher; curl -sL https://git.io/fisher | source && fisher update || fisher install jorgebucaran/fisher; end
+```
+
 ## Quickstart
 
 You can install, update, and remove plugins interactively with Fisher, taking advantage of Fish [tab completion](https://fishshell.com/docs/current/index.html#completion) and rich syntax highlighting.


functions/fisher.fish
```diff
@@ -82,7 +82,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                 command mkdir -p $source/{completions,conf.d,themes,functions}
 
-                $fish_path --command "
+                $fish_path --no-config --command "
```
jorgebucaran, line 85: What does --no-config do?

Susensio: As stated in #748 it avoids recursion, making easier to write confs files (not having to test for `status is-interactive`)
I've checked the command after line 85 and it seems to work fine with fish without config, as it should not require any user customizations
In addition, not reading config when it's not necessary should result in faster execution.
This `--no-config` flag was added on https://github.com/fish-shell/fish-shell/pull/7921 in order to fix https://github.com/fish-shell/fish-shell/issues/5394, that is, precisely avoid guarding with `status --is-interactive`

jorgebucaran: > faster execution

Awesome. 💯 

If we start using --no-config (and I think we should) we don't need to add this to the docs.

jorgebucaran: @Susensio Could you open a new PR that only includes the --no-config fix? Or modify this one like so. You are welcome to open a new PR to improve the docs if you want.

jorgebucaran (changes requested):

functions/fisher.fish
```diff
@@ -2,6 +2,8 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
     set --query fisher_path || set --local fisher_path $__fish_config_dir
     set --local fisher_version 4.4.3
     set --local fish_plugins $__fish_config_dir/fish_plugins
+    test -e $fish_plugins && set --local file_plugins (string match --regex -- '^[^\s]+$' <$fish_plugins)
+    set --query _fisher_plugins || set --universal _fisher_plugins $file_plugins
```
jorgebucaran, line 6: @Susensio What is this change about?

jorgebucaran (changes requested):

functions/fisher.fish
Susensio: This is a merge from #746 that I added into my personal fork, I did not know it would modify this PR

Susensio: Let me split this into multiple PRs so you can decide better

Susensio: The --no-config part is no longer needed. I thought that it was intended to avoid `status is-interactive` but it is more than that.
The subshell fisher starts to parallelize installation does not need to read config, as it seem (at the moment) to only write to temp folders and such. It wouldn't hurt to add this, but the improvement in speed is negligible so I think it's safer and simpler to **let it be as is**.

Susensio:
Simplified and fixed in #756

PR #756, opened 2023-01-21
Document autoinstall (bootstrap)

Susensio:
This started in #748, and is a partial PR from #749.
Instead of directly doing a `fisher install jorgebucaran/fisher`, it is recommended to try to `fisher update` first. This honors a maybe presente `fish_plugins` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@ Manage functions, completions, bindings, and snippets from the command line. Ext
 curl -sL https://git.io/fisher | source && fisher install jorgebucaran/fisher
 ```
 
+### Automatic install
+
+If you want to automatically install fisher when it is not present, you can add
+the following at the top of your `~/.config/fish/config.fish`.
+If `fish_plugins` is present is readed first.
+
+```fish
+if not functions -q fisher; curl -sL https://git.io/fisher | source && fisher update || fisher install jorgebucaran/fisher; end
+```
+
 ## Quickstart
 
 You can install, update, and remove plugins interactively with Fisher, taking advantage of Fish [tab completion](https://fishshell.com/docs/current/index.html#completion) and rich syntax highlighting.
--- a/README.md
+++ b/README.md
@@ -24,7 +24,9 @@ the following at the top of your `~/.config/fish/config.fish`.
 If `fish_plugins` is present is readed first.
 
 ```fish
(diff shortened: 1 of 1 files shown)


jorgebucaran:
@Susensio Do you no longer recommend using `--no-config`?

Susensio:
I think it does **not** add anything that the plugin needs, so I have no reason to use it. The small improvement in speed is not worth the future problems that it could cause in the future in case that config does something important.

README.md
```diff
@@ -17,6 +17,18 @@ Manage functions, completions, bindings, and snippets from the command line. Ext
 curl -sL https://git.io/fisher | source && fisher install jorgebucaran/fisher
 ```
 
+### Automatic install
+
+If you want to automatically install fisher when it is not present, you can add
+the following at the top of your `~/.config/fish/config.fish`.
+If `fish_plugins` is present is readed first.
```
djpowers, line 24:
```suggestion
If `fish_plugins` is present, it is read first.
```

Susensio:
Just reviewing old PRs in my personal forks, why was this closed @jorgebucaran ? I thought it added value.

jorgebucaran:
I am happy to revisit this, but I don't think an automatic installation workflow belongs in the core project's README. I don't have a perfect alternative location to suggest either. Sorry. Don't this kind of knowledge really belong in a user's (your) personal dotfiles?

Susensio:
Fair point. It is in my dotfiles indeed. I've always loved to have my dotfiles declarative, and I find very useful when tools propose ways of autoinstalling. If not, I have to remember to install them on a fresh system, check the docs or write some small script. I would rather have it already resolved in my config.

Besides, these conditional auto installers are non obvious and it's nice to be able to copy & paste a battle tested solution.In this case, the update before install respects the `fish_plugins` file, and one has to find that bug and find a solution.


tmux plugin manager has this documented
https://github.com/tmux-plugins/tpm/blob/master/docs/automatic_tpm_installation.md
lazy nvim also has this
https://lazy.folke.io/installation
even fisher had in the past
https://github.com/jorgebucaran/fisher/blob/dadcc6c19065ed06d43c6457718f2759964f7680/README.md?plain=1#L109-L120

I'm sure people find these things useful.
Anyhow, if you feel like this is better left as is, let it be.

PR #759, opened 2023-02-11
Fix fishshell.com configuration link anchor

jhillyerd:
Fish changed their docs, `configuration-files` anchor no longer exists, but `configuration` has a decent intro.


jorgebucaran:
Thank you, @jhillyerd! 💯

PR #766, opened 2023-07-05
prepend `command` to curl and tar

razvanazamfirei:
This is a pretty trivial change. 

The PR adds the `command` prefix in front of the `curl` and `tar` commands to avoid alias interference.


jorgebucaran:
Thanks, @razvanazamfirei! 💯

PR #771, opened 2023-08-19
fix: ignore .curlrc file when downloading plugin's tarball

ZuBB:
When `.curlrc` contains `-O` option, curl saves file to disk instead of redirecting it to stdout. thats why tar fails at next step as there is noting to extract

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -98,7 +98,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
 
-                        if command curl --disable --silent --location \$url | command tar -xzC \$temp -f - 2>/dev/null
+                        if command curl --disable --silent -L \$url | command tar -xzC \$temp -f - 2>/dev/null
                             command cp -Rf \$temp/*/* $source
                         else
                             echo fisher: Invalid plugin name or host unavailable: \\\"$plugin\\\" >&2
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -98,7 +98,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
 
-                        if command curl --silent -L \$url | command tar -xzC \$temp -f - 2>/dev/null
+                        if command curl --disable --silent --location \$url | command tar -xzC \$temp -f - 2>/dev/null
                             command cp -Rf \$temp/*/* $source
                         else
                             echo fisher: Invalid plugin name or host unavailable: \\\"$plugin\\\" >&2
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -98,7 +98,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
 
                         echo Fetching (set_color --underline)\$url(set_color normal)
 
-                        if command curl --disable --silent -L \$url | command tar -xzC \$temp -f - 2>/dev/null
+                        if command curl -q --silent -L \$url | command tar -xzC \$temp -f - 2>/dev/null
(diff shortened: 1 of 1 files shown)


jorgebucaran:
This might mess things up or break stuff, but not sure how bad. Maybe we should go for something like `fisher_curl_flags` instead? Thoughts?

ZuBB:
A couple of thoughts related to your comment

> This might mess things up or break stuff

I am sorry, but between "might" and "already" I would choose to fix bugs that already exist. "might" may end as "nothing wrong happened".

What fisher needs is to download a file. Downloading a file (through HTTP proto) is a pretty routine operation within a system. As a rule, there are several tools/apps in a system that work with HTTP. So it is hardly imaginable that each of that apps has its own HTTP settings that make download/"serve" actually work. And even if that is the case, I would say that particular system/setup is more screwed than mine with `-O` in `.curlrc`.

> Maybe we should go for something like fisher_curl_flags instead?

I am not sure I get your idea. can you give a bit more details?

ZuBB:
> This might mess things up or break stuff, but not sure how bad

on the contrary: can you come up with the craziest example of the harm that happens when a fisher tries to download a file for itself with the help of curl and not using the user's `.curlrc` file?

jorgebucaran:
Can't think of one yet. Let's skip .curlrc then. For now, stick to the single-letter version of the flag. We can switch to the long version in a separate commit later.

ZuBB:
Should I replace all flags with their single letter versions?

ZuBB:
`-qsl` or `-q -s -l`?

jorgebucaran:
No, only the ones you've made changes to. What's the short version of --disable?

jorgebucaran:
We can go for the longer versions later. I just want to keep this commit separate.

ZuBB:
> No, only the ones you've made changes to. What's the short version of --disable?

`-q`

ZuBB:
Sorry, I am not sure what to do here. Replace "disable" with "q"?

jorgebucaran:
Why did you change -L for --location?

ZuBB:
I just thought that it would be better (more descriptive) to have the full version of the flags instead of a short one. I am OK with either. Should I put `-L` back?

jorgebucaran:
Yes

jorgebucaran:
Use the long one. I'll be using long ones for all flags later on, but I want to keep this commit separate.

ZuBB:
Done. If you need it squashed -- let me know

jorgebucaran:
Hey, thinking it over, let's roll with `-q`. I was checking out the curl 7.49.0 [release notes](https://curl.se/changes.html#7_49_0) and saw:

> curl: make --disable work as long form of -q

Looks like `-q` was around before `--disable`, and some users might still be on an earlier curl.

ZuBB:
Done

ZuBB:
I have a bit off-top q: what happens with the content of `fisher_plugins` files when I run the install command? For example, I need to bootstrap my env after moving to a new machine and the subject file arrived through cloned dotfiles. plain overwrite or any kind of smart action(s)?

jorgebucaran:
If you've moved to a new machine and cloned your dotfiles, running `fisher install` should set up your plugins as per your previous machine, using the `fisher_plugins` file as a reference.

jorgebucaran:
Well, to bootstrap, you first install Fisher using our installation script:

```console
curl -sL https://raw.githubusercontent.com/jorgebucaran/fisher/main/functions/fisher.fish | source && fisher install jorgebucaran/fisher
```

The script directly sources and then installs Fisher from GitHub. It does this because Fisher isn't already in your `~/.config/fish/functions` or its universal state.

For plugin updates, Fisher replaces old files with new ones and removes any outdated files from the plugin.

ZuBB:
thank you for the response. and thanks a lot for your work as fisher's maintainer

Issue #624, opened 2020-11-17
Plugin installation can clobber existing files


mattgreen:
Hi,
I had a user file an [issue](https://github.com/mattgreen/lucid.fish/issues/6) where they installed a package I wrote and then didn't realize it'd clobber their existing `fish_prompt.fish`. It'd be nice if Fisher warned users that it was about to overwrite an existing file that was outside the current package. (Though I realize the last part of that is non-trivial.)

jorgebucaran:
@mattgreen Maybe we can copy any existing prompt files somewhere, and restore them after you uninstall the new prompt? We already attempt to restore `fish_prompt` with the default Fish prompt here:

https://github.com/jorgebucaran/fisher/blob/4ff6128860500f9cc39e16803f045e89d2a4a648/fisher.fish#L165

jorgebucaran:
We could copy any existing prompt files somewhere, and restore them after you uninstall the prompt. We already attempt to restore `fish_prompt` with the default Fish prompt here:

https://github.com/jorgebucaran/fisher/blob/4ff6128860500f9cc39e16803f045e89d2a4a648/fisher.fish#L165

jorgebucaran:
This is a non-trivial issue. Asking the user _Overwrite "$file"?_ isn't great, as it would effectively only half-install the plugin if the user doesn't want to overwrite. We may also need to introduce a `--yes` or `--overwrite` flag, since people may be using Fisher in automated scripts.

Here are some of the options we have, all with their own cons and pros:

1) Do nothing. Users should be careful not to install plugins that could potentially overwrite other plugins. The most typical example of this is installing a prompt. What should we do about: `fisher install a/foobar b/foobar`. Now, prompt authors could provide a custom `*_install` event and deal with this themselves, but that's not ideal either.

2) Abort installing plugins that would overwrite files (that Fisher isn't already tracking) in `$fisher_path`. That means `a/foobar` gets installed above, but `b/foobar` isn't. It also means, neither gets installed if the user already had a `foobar.fish` in their `$fisher_path`. Print a good error message.

3) Same as (2), but instead of aborting, handle it somehow. For example, copy all would-be-overwritten files somewhere so we can restore them if the user `fisher remove PLUGIN`.

kopischke:
[...]

PR #774, opened 2023-08-29
fix: remove conflicting files when plugins aren't indexed

Adamekka:
This is an annoying issue in fisher: When you move to a new PC, the `fish_variables` file which includes paths what files, and what plugin isn't present and fisher doesn't want to overwrite those unindexed files, my PR removes those conflicting files and then everything works


jorgebucaran:
This setup exists because of https://github.com/jorgebucaran/fisher/issues/624, (see c8091b3dee86b669c03b9f457c72c6a2c8b1c485). If we merge this change, it'd be like undoing that. To resolve your issue, you can simply uninstall the conflicting plugin and then go ahead with your intended installation.

PR #780, opened 2023-11-15
Update fisher.fish to support codeberg.org

Shea690901:


jorgebucaran:
Dupe https://github.com/jorgebucaran/fisher/pull/737

PR #784, opened 2023-12-24
Rename jorgebucaran -> joxji in files

### LINKED ISSUE #783: Update project with username change
Seems like the maintainer's username was changed from `jorgebucaran` to `joxji`, but the installation instructions in [README.md](./README.md) and other files in this project still reference the old username. GitHub redirects the old links, but this will break if someone else happens to register with the old username.

thesilican:
Resolves #783

By the way the url in `LICENSE.md` seems to be broken


joxji:
I'll likely just remove the link. I'm currently away from my desk, but I'll take care of it myself later. Thanks.

PR #792, opened 2024-09-09
Use '~' for $HOME in fish_plugins file rather than resolving path

mattmc3:
The fish_plugins file isn't portable across machines for local plugins because it stores the path to `$HOME`, which could be different on a different machine. This PR fixes this.

Steps to reproduce with Fisher 4.4.4:

```fish
$ fisher install $__fish_config_dir/plugins/foo
fisher install version 4.4.4
Installing /Users/mattmc3/.config/fish/plugins/foo
           /Users/mattmc3/.config/fish/.fisher/functions/foo.fish
Updated 1 plugin/s
$ cat $__fish_config_dir/fish_plugins
jorgebucaran/fisher
/Users/mattmc3/.config/fish/plugins/foo
```

This PR for Fisher 4.4.5, which makes fish_plugins more portable across machines with alternative user names:

```fish
$ fisher install $__fish_config_dir/plugins/foo
fisher install version 4.4.5
Installing /Users/matt/.config/fish/plugins/foo
           /Users/matt/.config/fish/.fisher/functions/foo.fish
Updated 1 plugin/s
$ cat $__fish_config_dir/fish_plugins
jorgebucaran/fisher
~/.config/fish/plugins/foo
```

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -29,7 +29,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
             set --local old_plugins $_fisher_plugins
             set --local new_plugins
 
-            test -e $fish_plugins && set --local file_plugins (string match --regex -- '^[^\s]+$' <$fish_plugins)
+            test -e $fish_plugins && set --local file_plugins (string match --regex -- '^[^\s]+$' <$fish_plugins | string replace -- \~ ~)
 
             if ! set --query argv[2]
                 if test "$cmd" != update
@@ -206,7 +206,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                     contains -- (string lower -- $plugin) (string lower -- $commit_plugins) || set --append commit_plugins $plugin
                 end
 
-                printf "%s\n" $commit_plugins >$fish_plugins
+                string replace --regex -- $HOME \~ $commit_plugins >$fish_plugins
             else
                 set --erase _fisher_plugins
                 command rm -f $fish_plugins


functions/fisher.fish
```diff
@@ -206,7 +206,7 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                     contains -- (string lower -- $plugin) (string lower -- $commit_plugins) || set --append commit_plugins $plugin
                 end
 
-                printf "%s\n" $commit_plugins >$fish_plugins
+                printf "%s\n" $commit_plugins | string replace --regex -- '^'$HOME'/' '~/' >$fish_plugins
```
mattmc3, line 209: Oops - I should not have piped from printf. I should have just made this line into:

```
string replace --regex -- '^'$HOME'/' '~/' $commit_plugins >$fish_plugins
```

If there are other changes you need, I can incorporate this one too.

jorgebucaran: Yes, go ahead! And while you're at it, use this format for the regex to keep it consistent with the other parts of the file where we handle the ~ to $HOME conversion.

```console
string replace --regex -- $HOME \~
```

Isn't this good enough?

jorgebucaran (changes requested):
We missed this when we did the same for universal variables, so I really appreciate it. Thanks!

functions/fisher.fish
mattmc3: > Isn't this good enough?

It is. There's an edge case where `/User/foo` includes stuff from `/User/foobar` and that string replace would make `~bar`. That is highly unlikely to be an actual issue for anyone in the real world, so I'll make the change.

jorgebucaran: Thanks! I should look into it and update all the other instances then. Those small inconsistencies tend to add up, and after a while, one can't remember if they were intentional or not. 😂

jorgebucaran (changes requested):
Just one more little thing! 🙏

functions/fisher.fish
```diff
@@ -1,6 +1,6 @@
 function fisher --argument-names cmd --description "A plugin manager for Fish"
     set --query fisher_path || set --local fisher_path $__fish_config_dir
-    set --local fisher_version 4.4.4
+    set --local fisher_version 4.4.5
```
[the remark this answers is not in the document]
mattmc3: Sure thing! I made the changes and squashed everything down to one commit. Let me know if this is good to go. Thanks, as always, for the thoughtful feedback.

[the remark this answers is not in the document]
jorgebucaran: Yes, thanks! I'll take it from here. 💯

jorgebucaran (approved):

PR #798, opened 2025-04-16
never remove ~/.config/fish/fish_plugins

lengyijun:
Partial fix #770 #790

Issue #790, opened 2024-06-02
fisher update  not install plugin but remove fish_plugins file


taikulawo:
I sync fish config across machine by git.
But on another machine I run `fisher update`, fisher just delete `fish_plugins`
![image](https://github.com/jorgebucaran/fisher/assets/24750337/0d6d2d16-f587-4701-8b1f-eb5394499905)

jorgebucaran:
Hmm, I really can't help without more info. Maybe share your entire configuration and sync strategy, maybe in a Docker image or a repository.

taikulawo:
extract to fish/ then run `fisher update`
[fish.tar.gz](https://github.com/user-attachments/files/15524164/fish.tar.gz)

jorgebucaran:
Could you share the contents here instead?

taikulawo:
sure. fish_plugins file
```
jorgebucaran/fisher
patrickf1/fzf.fish
```
and install fisher

jorgebucaran:
Is that what's in your gzipped tar file? What I need is your complete configuration and sync setup. Without it, reproducing your state won't be possible anytime soon.

taikulawo:
`https://github.com/taikulawo/home` I clone (means sync)it on my homedir

jorgebucaran:
The only file I'd put in my dot files would be fish_plugins" similar to how you wouldn't sync node_modules.

jorgebucaran:
The only file I'd put in my dot files would be fish_plugins, similar to how you wouldn't sync node_modules.

taikulawo:
I dont understand, I'm wrong?

jorgebucaran:
I saw you're committing all Fish files in your repo, including plugins likely installed with Fisher. In that case, what would be the purpose of fish_plugins? This file serves as a manifest of your plugins and should be enough for syncing across different machines. You'd simply use `fisher install` to install the plugins instead of relying on git.

taikulawo:
`does fisher install` not extra argument can install all plugin in manifest?
```
$ fisher install
fisher: Not enough arguments for command: "install"
```

jorgebucaran:
Thanks for the clarification. You're right, it's `fisher update` (not `install`), as indicated in the docs. 👍

jorgebucaran:
Yes, Fisher will delete fish_plugins, but only if it's empty. For example, if you remove all plugins from fish_plugins and then run `fisher update`, Fisher will uninstall all plugins and subsequently delete the file.

[...]

Issue #796, opened 2025-03-25
Allow Multiple Plugin Files


zchrykng:
Having multiple plugin files would enable having one checked into git and shared across machines, while having another that is local to the machine.

Mostly running into this need trying to get my fish config into a repo and sharing it between macOS and Linux, as I have plugins that are macOS only.

Edit: preference would be allowing something like fish_plugins_local or local/fish_plugins. Honestly, I'd like to normalize, having a local/ directory that mirrors the parent directory but allows local overriding of global configuration.

jorgebucaran:
Thanks for the suggestion, but this feature request isn't something I'm keen on adding. Maybe you can keep using fish_plugins as usual and for OS-specific plugins, install them conditionally in your config.fish _or_ enable the desired fish_plugins based on your OS as well. This keeps things simple here and avoids complexity.

zchrykng:
Okay, totally understand.

Is there guidance somewhere on how to accomplish that conditional installation, etc? I'm fairly new to fish and am not sure how to accomplish any of that. If it was zsh, I'd just check for the existence of a local file and source it if it exists, but not sure how I'd do something similar in fish.

PR #799, opened 2025-04-17
never remove ~/.config/fish/fish_plugins when update/install

### LINKED ISSUE #770: Feature request: option to force overwrite existing files
Hello @jorgebucaran,
long time no see!

I'm working to add [k8s support in Fish](https://github.com/pure-fish/pure/issues/322), when I try to install pure from my local directory, I'm met with a bunch of warning:

```
❯ fisher install ./                                                                                      
fisher install version 4.4.3                                                                             
fisher: Cannot install "/home/edouard/projects/contributions/pure": please remove or move conflicting files first:
        /home/edouard/.config/fish/functions/_pure_check_for_new_release.fish                            
        /home/edouard/.config/fish/functions/_pure_detect_container_by_cgroup_method.fish                                                                                                                          
        /home/edouard/.config/fish/functions/_pure_detect_container_by_pid_method.fish                   
        /home/edouard/.config/fish/functions/_pure_format_time.fish                                      
        /home/edouard/.config/fish/functions/_pure_get_prompt_symbol.fish         
``` 
and
```
❯ fisher install ./
fisher install version 4.4.3
fisher: Cannot install "/home/edouard/projects/contributions/pure": please remove or move conflicting files first:
        /home/edouard/.config/fish/functions/fish_greeting.fish
        /home/edouard/.config/fish/functions/fish_mode_prompt.fish
        /home/edouard/.config/fish/functions/fish_prompt.fish
        /home/edouard/.config/fish/functions/fish_title.fish
```

## Actual

The current solution is to run:
1. my [uninstall script `_pure_uninstall`](https://github.com/pure-fish/pure/blob/8de462db6274dcb13b46b2b2d2b930c086a3c9ec/conf.d/_pure_init.fish) 
2. then remove `rm $__fish_config_dir/functions/fish_*.fish`
3. finally install with `fisher install ./`

## Expected

```
fisher install --force ./
```

lengyijun:
Follow #798

Partial fix #770 #790

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -209,7 +209,9 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                 string replace --regex -- $HOME \~ $commit_plugins >$fish_plugins
             else
                 set --erase _fisher_plugins
-                command rm -f $fish_plugins
+		if test "$cmd" = remove
+		    command rm -f $fish_plugins
+		end
             end
 
             set --local total (count $install_plugins) (count $update_plugins) (count $remove_plugins)


lengyijun:
in #798, `fisher update` removes `fish_plugins`, if `fish_variables` is broken.

With this pr,  `fisher update`  will not remove `fish_plugins`.
To fix the broken `fish_variables`, we can 
```
fisher_path=(mktemp -d) fisher update
fisher update
```

jimblue:
Just tested, the fix is working great !

jorgebucaran:
The original code is saying that if `_fisher_plugins` is empty then it removes `fish_plugins` entirely to avoid leaving an empty plugins file behind. 

Your proposal makes sense too, but I don't understand why we need it or how it relates to #796. 

Can you clarify?

lengyijun:
> The original code is saying that if _fisher_plugins is empty then it removes fish_plugins entirely to avoid leaving an empty plugins file behind.

 _fisher_plugins is empty doesn't mean `fish_plugins` file is empty.
 _fisher_plugins is broken because of some accidents, for example, reset themes.
rm  `fish_plugins` means we forget all plugins we used to install

lengyijun:
796 should be a typo

jorgebucaran:
@lengyijun 
Do you mean there are cases where `_fisher_plugins` exists but `fish_plugins` does _not_, and that this only happens when `_fisher_plugins` is corrupted? If so, how does it get corrupted? If I can reproduce it, maybe we can prevent it.

lengyijun:
> how does it get corrupted

Not related to fisher.
`fish_variables` are removed to reset themes or reset environment variables


Most plugins don't rely on `fish_variables`.
I often remove `fish_variables`  to reset everything

lengyijun:
## Another case where `_fisher_plugins` is broken but `fish_plugins` is not: 

https://github.com/jorgebucaran/fisher/issues/790#issuecomment-2808244812

With this pr, `fish_plugins` is kept, that is important

jorgebucaran:
Why are you messing with `fish_variables` again?

lengyijun:
This isn't just my problem; many people have encountered this.

zchrykng:
> Why are you messing with `fish_variables` again?

People have been reporting this issue to you for years, maybe just accept that it happens to people on a semi-regular basis? There are lots of different things that can cause fish_variables to get changed or messed up and fisher deciding that a missing value in that file is grounds to delete the file that is *supposed* to be specifying what plugins should be installed is just crazy pants.

I don't know why plugins are stored in fish_variables, but I think what everyone wants is for the actual plugins file to be the source of truth rather than a variable in a file that isn't meant to be human readable or editable and can't be synced between machines because of environment differences.

I was checking out this project again after more than a year to see if things were fixed and sad to see that they still are not. This was the bug that stopped me using fisher ages ago at this point. I'd like to use a fish plugin manager, but have more or less given up at this point because of things like this.

jorgebucaran:
Have they? Do they? You seem to know more concrete facts about Fisher than I do. I've never personally encountered corrupted `fish_variables`. I've seen reports of it, yes, but it's by no means a semi-regular occurrence. That said, my goal here is to fix the underlying issue, not just treat the symptoms.

Let me clarify something that may have been misunderstood. When I said, "Why *were* you messing with `fish_variables` again?", I meant it as "Can you remind me why you were touching `fish_variables`?". Not in a condescending way like "Why are you messing with that again?". 

Obviously, `fish_variables` is an internal Fish file and shouldn't be modified directly. That said, I can see how people might end up doing so inadvertently. Not everyone has the patience to fully configure their shell, especially when coming from another shell, and people just want things to work so they can focus on their actual projects and jobs. Still, that's not really an excuse. It's a bit like manually corrupting your system memory and then blaming the software when things break.

The reason I'm hesitant to accept this patch right away and why I asked that question is that I want to understand exactly what caused this specific instance of `fish_variables` corruption. Once that's clear, we can think about how to properly fix or mitigate the issue, or at least make a better-informed decision. Without that understanding, this PR feels more like treating surface-level symptoms (like taking painkillers) rather than identifying and addressing the root cause, which is what I actually care about.

Plugins themselves are not stored in `fish_variables`, they live in the filesystem. Fisher does use universal variables to track plugin locations, and those universal variables are stored in `fish_variables`, which is a Fish-managed file. Fisher has gone through many iterations, we've done this differently in the past as well. We've had it the other way too. This architecture works better for Fisher right now.

The real issue here is that I haven't had the time to dig deeper yet (but I do now), and as a result I still don't understand the root cause of the issue. I could merge this PR and see what happens, but I'd prefer to understand the underlying problem first so I can propose a better solution.

zchrykng:
Apologies for coming in fairly hot in my first message. It did feel rather dismissive or condescending and I'm sorry if I misunderstood.

> I meant it as "Can you remind me why you were touching fish_variables?".

We generally aren't directly messing with it. Mismatches occur when syncing between systems because fish_variables can't be included in git without a ton of noise and manual merges. Syncing it causes potential problems because systems need/have different state and fish decided to stick a ton of stuff into that file rather than putting it in some cache file somewhere. It can also get messed with if you have decided to clear your fish state and start over. Since fish stores settings in there that aren't specified in configuration files, sometimes the only way to reset things is to delete the file.

What I think is happening, while admitting that I haven't gone through the fisher code myself at this point, is that fisher looks for the variables in fish_variables and when it doesn't find them, it deletes the plugins file rather than treating the plugins file as the source of truth and installing the plugins.

zchrykng:
Here is an example of what happens.

image

This is on a machine that I never used fisher on but I just loaded in my dotfiles. I ran the script to install fisher and it nukes the fish_plugins file.

jorgebucaran:
I can see how this could happen in this specific scenario, and yes, it could be briefly confusing. In that case, you should run:

```fish
curl -sL https://raw.githubusercontent.com/jorgebucaran/fisher/main/functions/fisher.fish | source && fisher update
```

since you already have a `fish_plugins` file. That said, I agree it would be nicer if Fisher could detect this automatically.

Now, this PR does _not_ address that issue, so I say let's tackle it separately.

jorgebucaran:
Tracking https://github.com/jorgebucaran/fisher/pull/799#issuecomment-3792239128 in #823

PR #806, opened 2025-08-07
Update tide example to v6, which has been out since 2023

technicalpickles:
I only realized this myself this past week. May as well update this example so people using it get the latest!


jorgebucaran:
Thank you, and sorry for ignoring this earlier. It was a no-brainer, and it took me forever, but I'm back.

PR #807, opened 2025-08-07
Add uninstall as an alias to remove

technicalpickles:
Possible implementation of https://github.com/jorgebucaran/fisher/issues/805

Some considerations:

- should the alias be included in the help output or not? only list is, but not ls
- should uninstall be the default instead of remove?

PR #808, opened 2025-08-07
Add completions for installed plugins 

technicalpickles:
I noticed this while working on https://github.com/jorgebucaran/fisher/pull/807 . I figured it'd be nice to be able to complete the name of installed plugins for remove and update.


jorgebucaran:
Thanks for your patience, I'm just taking a look at this PR. The fisher update and fisher remove subcommands already automatically complete valid arguments using fisher list.

https://github.com/jorgebucaran/fisher/blob/791da644d33d392216f6b1a9b5fc1e470db6d7f2/completions/fisher.fish#L7

It seems you noticed this, so why propose a refactor? We can continue using fisher list for free instead of adding a new function.

jorgebucaran:
Closing.

PR #811, opened 2025-09-30
Support arbitrary git hosts

### LINKED ISSUE #801: Feature request: support git.sr.ht
I'd like to publish a Fish plugin, but I host my projects on sr.ht. It seems like `fisher` only supports GitHub and GitLab. Would be nice to add support for git.sr.ht as well!

furudean:
Add support to fetch plugins from arbitrary git hosts via `git clone`. It will use whatever the user's git config is to clone things.

These patterns are now possible:
```
fisher install https://git.sr.ht/~jlkde/note.fish
fisher install git@tangled.sh:himawari.fun/bugstyle.fish
fisher install git@github.com:meaningful-ooo/sponge.git@1.1.0
```

We assume that any URL that doesn't match existing officially supported git hosts like github `user/repo(@ref)?` or `(https?://)?gitlab.com/*` is a git host.

I have tested this with these plugins:
```
jethrokuan/z
https://github.com/jorgebucaran/spark.fish.git
gitlab.com/mirdono/tucker
git@github.com:meaningful-ooo/sponge.git@1.1.0
git@tangled.sh:himawari.fun/bugstyle.fish
https://git.sr.ht/~jlkde/note.fish
```

Fixes #801
Maybe fixes #745 as requests can be authenticated using ssh

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -226,6 +226,13 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
                 contains -- $plugin $install_plugins && set --local event install || set --local event update
 
                 printf "%s\n" Installing\ (set_color --bold)$plugin(set_color normal) "           "(string replace --regex -- '^'$HOME '~' $$plugin_files_var)
+
+                for file in (string match --regex -- '.+/[^/]+\.fish$' $$plugin_files_var | string replace --regex -- '^'$HOME '~')
+                    source $file
+                    if set --local name (string replace --regex -- '.+conf\.d/([^/]+)\.fish$' '$1' $file)
+                        emit {$name}_$event
+                    end
+                end
             end
 
             command rm -rf $source_plugins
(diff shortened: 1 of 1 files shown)


bdmorin:
I just took a stab at this, didn't see your PR, thank you!

jorgebucaran:
Thanks, but I don't want to invest resources in maintaining these patterns. Fisher currently works best with GitHub.

leifssm:
first, thank you for fisher @jorgebucaran! second, while i understand the argument, these patterns arent likely to change as a lot more than just fisher depends on the fact that these patterns are consistent. if they work now there is a very low chance that it will break in the future, so please reconsider

jorgebucaran:
Instead of supporting any URL, we could just limit it to GitHub, GitLab, and Sourcehut. 

That good enough, or do you really need to support any arbitrary Git host?

PR #822, opened 2026-01-24
ignore same files

lengyijun:


jorgebucaran:
The current behavior in Fisher is that if a user has a file with the same name as one in the plugin being installed, the install is aborted and the conflict is reported. This proposal suggests treating a file as a conflict only if its contents differ, allowing the installation to proceed silently otherwise. Smart, but it introduces a problem. Consider this scenario:

```console
$ fisher ls
jorgebucaran/fisher
/Users/me/local/foobar
```

With the proposed behavior, installing a remote version of foobar would overwrite the local one:

```console
$ fisher install remote/foobar
fisher install version 4.4.6
Fetching https://api.github.com/repos/remote/foobar/tarball/HEAD
Installing remote/foobar
           /Users/me/.config/fish/functions/foobar.fish
```

Now Fisher believes two versions of foobar exist:

```console
$ fisher ls
jorgebucaran/fisher
/Users/me/local/foobar
remote/foobar
```

This isn't really true, is it? Instead, Fisher should silently uninstall the conflicting version so that its internal state reflects the actual installed version.

Now, let me say, this is very clever. Nice work.

Also, `cmp` appears to be compatible and available across all Unix systems, so unless further investigation uncovers an issue, we should be good on that.

jorgebucaran:
Let's review these three scenarios:

#### User has foo.fish and installs a plugin with foo.fish:

- Before: Blocked (must remove manually)
- After: Proceeds if identical

But how did the file end up identical? Either they copied it from the plugin or it's an impossible coincidence. This is the only case that isn't harmful, because if the files are identical, no harm is done. Fisher can manage it now, which is a positive. 

#### User has plugin A and installs plugin B that shares files:

- Before: Blocked — correct, two plugins shouldn't share files
- After: Proceeds if identical — wrong, Fisher now thinks both own the file

#### User has a local plugin and installs a remote version:

- Before: Blocked — they're trying to replace the local version
- After: Proceeds but breaks state 

If anything, Fisher should update or remove then install, which is what I suggested in https://github.com/jorgebucaran/fisher/pull/822#issuecomment-3793974609.

In normal usage, matching files either mean the plugin is already installed or the user should be updating, not installing anew.

lengyijun:
I need to check if belongs to other plugin

jorgebucaran:
Yes, but even if you did that, I'm not sure what it would really buy us here. Feel free to think more about it, but I'm not going to keep this open for now. Closing.

PR #830, opened 2026-03-16
Make "fisher update" only update changed plugins

snoack:
With this change `fisher update` still re-downloads all installed plugins (as it used to) but skips re-installing them if the contents are unchanged which most importantly silences noise in the `fisher update` output about unchanged plugins being updated.


illutr0n:
Release it?

snoack:
Yeah, that would be great. @jorgebucaran

PR #836, opened 2026-07-31
Use `gh` or `GITHUB_TOKEN` if available to authenticate to private repos

### LINKED ISSUE #835: Unable to install a private plugin
I've created a private fisher plugin repository at https://github.com/github/offbyone-fish-shell-extensions

It's valid, I can access it with my `git` and `gh`: 

```shellsession
$ gh api repos/github/offbyone-fish-shell-extensions  | jq .url
"https://api.github.com/repos/github/offbyone-fish-shell-extensions"
```

However, fisher won't install it:

```shellsession
fisher install version 4.4.8
Fetching https://api.github.com/repos/github/offbyone-fish-shell-extensions/tarball/HEAD
fisher: Invalid plugin name or host unavailable: "github/offbyone-fish-shell-extensions"
$ fisher install github.com/github/offbyone-fish-shell-extensions
```

offbyone:
Fixes https://github.com/jorgebucaran/fisher/issues/835 which I noticed while trying to work with my work-private fish extensions

### DIFF
--- a/functions/fisher.fish
+++ b/functions/fisher.fish
@@ -8,8 +8,8 @@ function fisher --argument-names cmd --description "A plugin manager for Fish"
             echo "fisher, version $fisher_version"
         case "" -h --help
             echo "Usage: fisher install   <plugins...>  Install plugins"
-            echo "       fisher remove    <plugins...>  Remove installed plugins" 
-            echo "       fisher uninstall <plugins...>  Remove installed plugins (alias)" 
+            echo "       fisher remove    <plugins...>  Remove installed plugins"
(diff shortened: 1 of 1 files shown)


jorgebucaran:
I don't want to use `gh`. That would make Fisher depend on it, and I want Fisher to remain lightweight.

offbyone:
How would you feel if the use was optional (ie, check for `gh` on path, and use it only if present)?

offbyone:
(I think, btw, that since fisher already implicitly depends on github by assuming that a `org/name` plugin is on github, depending on the `gh` CLI isn't really adding a significant conceptual dependency. I agree that fisher should not _require_ gh to operate)

